PR #2, opened 2025-02-13
Refactor optimizer into FunctionOptimizer class

alvin-r:


codeflash/optimization/optimizer.py
```diff
@@ -100,6 +47,22 @@ def __init__(self, args: Namespace) -> None:
 
         self.test_files = TestFiles(test_files=[])
 
+    def create_function_optimizer(
```
misrasaurabh1, line 50: this should not belong in the Optimizer class but it is a concern of the FunctionOptimizer class. You can either do the validation in the init fuction or implement a classmethod within FunctionOptimizer

misrasaurabh1 (approved):

codeflash/optimization/optimizer.py
alvin-r: good catch

alvin-r: I'll merge this change for now, and rework this later.

alvin-r:
redoing this in a new PR

PR #3, opened 2025-02-14
compare stdout

KRRT7:


codeflash/verification/parse_test_output.py
```diff
@@ -278,9 +282,10 @@ def parse_test_xml(
                         test_type=test_type,
                         return_value=None,
                         timed_out=timed_out,
+                        stdout=stdout,
```
misrasaurabh1, line 285: this should be the testcase.system_out, with the r"!######(.*?):(.*?)([^\.:]*?):(.*?):(.*?):(.*?)######!" part cleaned out. It cannot be the global stdout, it has a lot of other things. This should just be the stdout print when for a particular FunctionTestInvocation

PR #10, opened 2025-02-15
Function optimizer refactor

alvin-r:

PR #13, opened 2025-02-17
go fish

KRRT7:
previously methods weren't parsed properly showing a 0% coverage metric. this is a bug fix for that.


KRRT7:
implemented in #10 instead

PR #16, opened 2025-02-17
Install black/ruff if not found in current python environment CF-495

aseembits93:
Added a try/except block which tries to import the formatter, it calls `pip install` via `subprocess` if the import fails.

### DIFF
--- a/codeflash/LICENSE
+++ b/codeflash/LICENSE
@@ -3,7 +3,7 @@ Business Source License 1.1
 Parameters
 
 Licensor:             CodeFlash Inc.
-Licensed Work:        Codeflash Client version 0.9.x
+Licensed Work:        Codeflash Client version 0.0.x
                       The Licensed Work is (c) 2024 CodeFlash Inc.
 
 Additional Use Grant: None. Production use of the Licensed Work is only permitted
@@ -13,7 +13,7 @@ Additional Use Grant: None. Production use of the Licensed Work is only permitte
                       Platform. Please visit codeflash.ai for further
                       information.
 
-Change Date:          2029-01-06
+Change Date:          2029-02-17
 
 Change License:       MIT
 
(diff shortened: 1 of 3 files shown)


codeflash/cli_cmds/cmd_init.py
```diff
@@ -578,8 +578,20 @@ def configure_pyproject_toml(setup_info: SetupInfo) -> None:
     formatter_cmds = []
     if formatter == "black":
         formatter_cmds.append("black $file")
+        print("checking if black exists")
+        try:
+            import black
```
misrasaurabh1, line 583: we should do this checking exactly the same way we run black when we do the formatting. I believe it runs as a subprocess. We should detect with that method if black is installed. The same goes for ruff formatter

codeflash/cli_cmds/cmd_init.py
```diff
@@ -578,8 +578,20 @@ def configure_pyproject_toml(setup_info: SetupInfo) -> None:
...
+        except ImportError as e:
+            print("black not found, installing via pip")
+            subprocess.run(["pip", "install", "black"])
```
misrasaurabh1, line 586: we should not modify the user environment like this, only recommend the user to install black with this command. same for ruff

codeflash/LICENSE
```diff
@@ -3,7 +3,7 @@ Business Source License 1.1
 Parameters
 
 Licensor:             CodeFlash Inc.
-Licensed Work:        Codeflash Client version 0.9.x
+Licensed Work:        Codeflash Client version 0.0.x
```
misrasaurabh1, line 6: also this file should be reverted

codeflash/version.py
```diff
@@ -1,3 +1,3 @@
 # These version placeholders will be replaced by poetry-dynamic-versioning during `poetry build`.
-__version__ = "0.9.2"
-__version_tuple__ = (0, 9, 2)
+__version__ = "0.0.0.post2491.dev0+2f81f2ca"
```
misrasaurabh1, line 2: this file should also be reverted.

PR #25, opened 2025-02-24
Don't display result in report table if INIT_STATE_TEST

davidgirdwood1:


codeflash/github/PrComment.py
```diff
@@ -20,6 +20,13 @@ class PrComment:
     winning_benchmarking_test_results: TestResults
 
     def to_json(self) -> dict[str, Union[dict[str, dict[str, int]], int, str]]:
+
+        report_table = {
+            test_type.to_name(): result
+            for test_type, result in self.winning_behavioral_test_results.get_test_pass_fail_report_by_type().items()
+            if test_type.should_display_result()
+        }
+
```
KRRT7, line 29: we don't need the extra method here, we can just inline it in the comprehension directly. i.e
```py
        report_table = {
            test_type.to_name(): result
            for test_type, result in self.winning_behavioral_test_results.get_test_pass_fail_report_by_type().items()
            if test_type.to_name()
        }

KRRT7 (changes requested):

KRRT7 (approved):
LGTM!

PR #34, opened 2025-03-01
ruff check & format

KRRT7:


misrasaurabh1 (approved):
thanks! the code looks prettier!

PR #40, opened 2025-03-06
[CF-551] Fix extra test type in TestType

dasarchan:
Removed TestType.INIT_STATE_TEST from TestType


misrasaurabh1 (approved):
Looks great!

PR #42, opened 2025-03-06
reworked exception handling

alvin-r:
reworked exceptions to not be raised by default. any 'with pytest' should also be removed, since it's already captured.


alvin-r:
will pause this for now. bugs were caused by the coverage issue, not exceptions

PR #44, opened 2025-03-10
Multi-threading support for the tracer

KRRT7:


codeflash/tracer.py
```diff
@@ -57,14 +85,7 @@ def __init__(
...
     ) -> None:
-        """:param output: The path to the output trace file
-        :param functions: List of functions to trace. If None, trace all functions
-        :param disable: Disable the tracer if True
-        :param config_file_path: Path to the pyproject.toml file, if None then it will be auto-discovered
-        :param max_function_count: Maximum number of times to trace one function
-        :param timeout: Timeout in seconds for the tracer, if the traced code takes more than this time, then tracing
-                    stops and normal execution continues. If this is None then no timeout applies
-        """
+        """Initialize Tracer."""
```
misrasaurabh1, line 88: dont remove the docstring

testbench.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def add_numbers(a: int, b: int) -> int:
+    print(f"[ADD_NUMBERS] Starting with parameters: a={a}, b={b}")
+    result = a + b
+    print(f"[ADD_NUMBERS] Returning result: {result}")
+    return result
+
+
+def test_threadpool() -> None:
```
misrasaurabh1, line 11: please integrate this test bench functionality into the tracer e2e test we already have

misrasaurabh1:
reviewed this async - main feedback being that tracer has to not create new sqlite db for each therad. write to the single sqlite db concurrently. Sqlite should handle the locking etc, thats what it is meant for.
Also avoid modifying the profiler code as we want to minimize the difference wrt the original python profile implementation.

.github/workflows/end-to-end-test-tracer-replay_testbench.yaml
```diff
@@ -0,0 +1,41 @@
+name: end-to-end-test
```
misrasaurabh1, line 1: we don't need 2 e2e tests for tracer. Modify the original tracer replay test to also incorporate the multi threaded code tracing

codeflash/tracer.py
```diff
@@ -127,29 +155,34 @@ def __enter__(self) -> None:
...
-    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
+    ) -> None:
         if self.disable:
             return
         sys.setprofile(None)
         self.con.commit()
-
+        console.rule("Program Output End", style="bold blue")
```
misrasaurabh1, line 185: there is no context provided to the user that this message is being printed by codeflash, confusing for the user.

KRRT7: adding
`Codeflash: Traced Program Output Begin` instead, makes it clearer to the user that the output is from the traced program itself

misrasaurabh1 (approved):

PR #45, opened 2025-03-10
fixed project config interfering with coverage config

alvin-r:
When parsing results, we use the Coverage class. we need to pass in the config file that we generate.


KRRT7 (changes requested):
is there a reason we're returning the coverage_config_file?

KRRT7 (approved):

PR #125, opened 2025-04-04
Better LLM Explanations CF-597

aseembits93:
wip

### DIFF
--- a/codeflash/optimization/function_optimizer.py
+++ b/codeflash/optimization/function_optimizer.py
@@ -261,6 +261,7 @@ class FunctionOptimizer:
                         best_optimization.candidate.explanation, title="Best Candidate Explanation", border_style="blue"
                     )
                 )
+                get_new_explanation = "abc"
                 explanation = Explanation(
                     raw_explanation_message=best_optimization.candidate.explanation,
                     winning_behavioral_test_results=best_optimization.winning_behavioral_test_results,


aseembits93:
# Create PR explanations through LLMs

Our current explanations suck, and need to be improved. This would make it easier for reviewers to accept our PRs.

- [ ] Create a new endpoint using litellm. This way we can easily try out any model, without any breaking change.
- [ ] for this new end point, from the cli, give it info that can help write better explanation
  - [ ] pre-existing explanation generated with llm
  - [ ] original code
  - [ ] new code 
  - [ ] timing info
  - [ ] maybe profiling info
- [ ] This should happen as the last step right before the PR is created. 
- [ ] The PR should be created with this info, replacing the current explanations
- [ ]

codeflash/api/aiservice.py
```diff
@@ -278,6 +278,76 @@ def optimize_python_code_refinement(self, request: list[AIServiceRefinerRequest]
...
+        except requests.exceptions.RequestException as e:
+            logger.exception(f"Error generating explanations: {e}")
+            ph("cli-optimize-error-caught", {"error": str(e)})
+            return ""
+
+        if response.status_code == 200:
+            explanation: str = response.json()["explanation"]
+            logger.info(f"New Explanation: {explanation}")
+            console.rule()
+            return explanation
```
KRRT7: look at how we currently display the explanation and then line this up with that, otherwise LGTM

KRRT7 (dismissed):

codeflash/api/aiservice.py
```diff
@@ -278,6 +278,76 @@ def optimize_python_code_refinement(self, request: list[AIServiceRefinerRequest]
...
+        try:
+            response = self.make_ai_service_request("/explain", payload=payload, timeout=600)
+        except requests.exceptions.RequestException as e:
+            logger.exception(f"Error generating explanations: {e}")
+            ph("cli-optimize-error-caught", {"error": str(e)})
+            return ""
+
+        if response.status_code == 200:
+            explanation: str = response.json()["explanation"]
+            logger.info(f"New Explanation: {explanation}")
```
misrasaurabh1: remove the info tag before merging

codeflash/api/aiservice.py
```diff
@@ -278,6 +278,76 @@ def optimize_python_code_refinement(self, request: list[AIServiceRefinerRequest]
...
+            "optimized_code_runtime": optimized_code_runtime,
+            "speedup": speedup,
+            "annotated_tests": annotated_tests,
+            "optimization_id": optimization_id,
+            "dependency_code": dependency_code,
+        }
+        logger.info("Generating explanation")
+        console.rule()
+        try:
+            response = self.make_ai_service_request("/explain", payload=payload, timeout=600)
```
misrasaurabh1: timeout of 600 is too much, make it 60. and if this request fails, but fallback not to empty string but the explanation we were using earlier.

codeflash/api/aiservice.py
```diff
@@ -278,6 +278,76 @@ def optimize_python_code_refinement(self, request: list[AIServiceRefinerRequest]
...
+
+        Returns
+        -------
+        - List[OptimizationCandidate]: A list of Optimization Candidates.
+
+        """
+        payload = {
+            "trace_id": trace_id,
+            "source_code": source_code,
+            "optimized_code": optimized_code,
```
misrasaurabh1, line 320: i think you should also include the original explanation, as at that time the "optimizer is speaking its mind". It would be easier to generate an accurate explantion with that info

codeflash/optimization/function_optimizer.py
```diff
@@ -1120,11 +1120,35 @@ def find_and_process_best_optimization(
                     )
                     if concolic_test_str:
                         generated_tests_str += "\n\n" + concolic_test_str
```
misrasaurabh1: make the delimiter as `\n#------------------------------------------------\n`

codeflash/api/aiservice.py
aseembits93: I'd argue that the explainer needs to just see the information needed to compare two code snippets and decide for itself in an unbiased manner how/why the new code snippet is faster using the line_prof/runtime information. I'm not sure how the original explanation could help with generating a new one?

aseembits93 (commented):
incorporated all the feedback

misrasaurabh1 (approved):

PR #129, opened 2025-04-07
Fix local setup path

HeshamHM28:


HeshamHM28:
After careful consideration, I decided to close this PR without making the fix due to the following reasons:

1. **Docusaurus Link Structure**: The link format used in the documentation (`/getting-started/local-installation`) is the standard structure recommended for Docusaurus. It works properly within the Docusaurus environment, and no issues exist there.
   
2. **GitHub Link Format Issue**: Although this format doesn't work in GitHub, modifying all the documentation links would create inconsistencies between how the documentation behaves on GitHub and within Docusaurus.

PR #211, opened 2025-05-16
Format only code that was optimized (and helpers)

zomglings:


codeflash/optimization/function_optimizer.py
```diff
@@ -581,25 +592,61 @@ def write_code_and_helpers(original_code: str, original_helper_code: dict[Path,
...
+                formatted_code,
+                self.function_to_optimize,
+                preexisting_functions_by_filepath[path],
+                helper_functions,
+            )
 
-        new_code = format_code(self.args.formatter_cmds, path)
-        if should_sort_imports:
-            new_code = sort_imports(new_code)
+            if len(code_ranges_formatted) != len(code_ranges_unformatted):
```
Saga4, line 626: Q: Can you help me understand the case on Why these have to be equal always?

codeflash/optimization/function_optimizer.py
```diff
@@ -299,7 +299,6 @@ def optimize_function(self) -> Result[BestOptimization, str]:
 
                 self.log_successful_optimization(explanation, generated_tests, exp_type)
 
-                # xylophone
```
misrasaurabh1: xylophone? 😂

misrasaurabh1:
approach looks good. Can you write tests for `reformat_code_and_helpers` that check for the variety of cases we expect around unformatted and formatted code?

codeflash/optimization/function_optimizer.py
zomglings: Easy to grep, should be gone. :)

zomglings:
No longer have permission to push to the official repo, will open this PR from my fork.

codeflash/optimization/function_optimizer.py
zomglings: Just a defensive measure. The function to optimize and helpers are non-overlapping blocks of code, and should not become overlapping due to formatting. If this has happened, something has gone wrong in the formatting step or in the extraction step (which identifies function to optimize and helpers).

PR #244, opened 2025-05-27
Ruff should target atleast py39

aseembits93:
This will ensure we do not accidentally reformat `Union` to `|`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -151,6 +151,7 @@ init_typed = true
 warn_required_dynamic_aliases = true
 
 [tool.ruff]
+target-version = "py39"
 line-length = 120
 fix = true
 show-fixes = true


pyproject.toml
```diff
@@ -151,6 +151,7 @@ init_typed = true
 warn_required_dynamic_aliases = true
 
 [tool.ruff]
+target-version = "py39"
```
Saga4, line 154: Q: Shouldn't this be an array as we are targeting py310, py311, py312 as well?

Saga4 (approved):

pyproject.toml
KRRT7: https://docs.astral.sh/ruff/settings/#target-version

it's the minimum version, so it's fine

PR #248, opened 2025-05-28
don't resolve directory when custom input is given for module root

KRRT7:


misrasaurabh1 (approved):

aseembits93 (changes requested):
some tests to check the behavior?

PR #206, opened 2025-05-14
Formatting should only target changes suggested by optimizer, not entire files (CF-637)

zomglings:
### **User description**
Closes CF-637

This PR implements targeted formatting for AI-generated code snippets within the `FunctionOptimizer`. The goal is to format only the specific function being optimized (and any helpers modified by the AI) rather than entire files with tools like Black/Ruff. Import sorting (`isort`) is still applied at a file level for correctness after the formatted snippet is integrated.

The existing `codeflash/code_utils/formatter.py` remains unchanged. All new temporary file management for string formatting is handled within `FunctionOptimizer`.

**Key Changes Implemented:**

*   **`FunctionOptimizer.__init__` Modifications:**
    *   [X] Imports `tempfile` (shutil and Path were already present).
    *   [X] Creates a base temporary directory for the optimizer instance: `self.optimizer_temp_dir = Path(tempfile.mkdtemp(prefix="codeflash_opt_fmt_"))`. This directory is *not* cleaned up by `FunctionOptimizer` itself, allowing it to persist for potential inspection or higher-level cleanup.

*   **`FunctionOptimizer.determine_best_candidate` Method Modifications:**
    *   [X] When `candidate.source_code` (a string) needs formatting (i.e., `self.args.formatter_cmds` is configured):
        *   [X] A temporary file is created within `self.optimizer_temp_dir`.
        *   [X] `candidate.source_code` is written to this temp file.
        *   [X] `format_code()` is called with the temp file's path. (`format_code()` handles the "disabled" formatter case internally).
        *   [X] The individual temporary file used for formatting is deleted in a `finally` block.
    *   [X] `candidate.source_code` is updated with the formatted snippet (output of primary formatter like Black/Ruff).
    *   [X] Import sorting on the snippet itself is *not* performed here; it's deferred to a file-level operation later.

*   **`FunctionOptimizer.optimize_function` Method Modifications:**
    *   [X] The `reformat_code_and_helpers` method and its call site are removed.
    *   [X] **New Import Sorting Logic**: After `replace_function_and_helpers_with_optimized_code` (which splices the formatted snippet into files):
        *   [X] If `not self.args.disable_imports_sorting`, `sort_imports()` is applied to the full content of the main function's file.
        *   [X] Similarly, `sort_imports()` is applied to the full content of any helper files (from `code_context.helper_functions`) that could have been modified.
[...]

PR #261, opened 2025-06-01
WIP - targeted function reformatting with CST

mohammedahmed18:
related #211 , #206
This PR focuses on formatting only the optimized (target) function, rather than the entire file, to minimize diff size.

> still requires some polishing and more unit tests to handle all corner cases.


mohammedahmed18:
Closing in favor of #274

PR #266, opened 2025-06-02
[Chore] warn message if the module root and test root are the same (CF-659)

mohammedahmed18:


codeflash/cli_cmds/cmd_init.py
```diff
@@ -72,7 +72,10 @@ def init_codeflash() -> None:
 
         if should_modify_pyproject_toml():
             setup_info: SetupInfo = collect_setup_info()
-
+            if Path(setup_info.module_root).resolve() == Path(setup_info.tests_root).resolve():
```
misrasaurabh1: do it when the user is actually entering the test directory, not after he is done.
Also remove the module_root path from the list of options when asking for the test_root. This will mostly solve the problem. We can still keep the warning in case the user enters the module_root path manually

misrasaurabh1 (approved):

PR #275, opened 2025-06-03
Don't repeatedly optimize gh actions

dasarchan:
Adds a call to cfapi to push hashes of the function code context in order to check if the function is being optimized again.

One thing to note here is that this checking logic happens early, in function discovery.

### DIFF
--- a/codeflash/discovery/functions_to_optimize.py
+++ b/codeflash/discovery/functions_to_optimize.py
@@ -15,14 +15,15 @@ import git
 import libcst as cst
 from pydantic.dataclasses import dataclass
 
-from codeflash.api.cfapi import get_blocklisted_functions
+from codeflash.api.cfapi import get_blocklisted_functions, make_cfapi_request
 from codeflash.cli_cmds.console import DEBUG_MODE, console, logger
 from codeflash.code_utils.code_utils import (
     is_class_defined_in_file,
     module_name_from_file_path,
     path_belongs_to_site_packages,
 )
-from codeflash.code_utils.git_utils import get_git_diff
+from codeflash.code_utils.env_utils import get_pr_number
+from codeflash.code_utils.git_utils import get_git_diff, get_repo_owner_and_name
 from codeflash.discovery.discover_unit_tests import discover_unit_tests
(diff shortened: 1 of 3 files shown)


codeflash/discovery/functions_to_optimize.py
```diff
@@ -417,6 +459,82 @@ def inspect_top_level_functions_or_methods(
...
+            code_contexts[path_key] = func_hash
+            path_to_function_map[path_key] = (file_path, func)
+
+    if not code_contexts:
+        return {}, 0
+
+    try:
+        # Call the optimization check API
+        logger.info("Checking status")
+        response = make_cfapi_request(
```
Saga4: For consistency, all the cf-api calls are directly happening through functions present in cfapi itself.
So we can just built a function for is-already-optimized there and call it using make_cf_api_request.

codeflash/discovery/functions_to_optimize.py
```diff
@@ -417,6 +459,82 @@ def inspect_top_level_functions_or_methods(
     )
 
 
+def check_optimization_status(
+        functions_by_file: dict[Path, list[FunctionToOptimize]],
```
Saga4: Q: What will be the scenario where the function is already an optimized code by CF?

dasarchan: Right now it would try to reoptimize it, actually - @misrasaurabh1 what's the desired behavior here

codeflash/discovery/functions_to_optimize.py
Saga4: @dasarchan

codeflash/optimization/function_optimizer.py
```diff
@@ -1280,7 +1299,7 @@ def generate_and_instrument_tests(
                 test_perf_path,
             )
             for test_index, (test_path, test_perf_path) in enumerate(
-                zip(generated_test_paths, generated_perf_test_paths)
+                zip(generated_test_paths, generated_perf_test_paths, strict=False)
```
misrasaurabh1: Revert this for 3.9

PR #289, opened 2025-06-05
give functionality to use codeflash:skip in label like other bots

Saga4:


misrasaurabh1:
@Saga4 how are users expected to use this?

Saga4:
> @Saga4 how are users expected to use this?

As roboflow asked to have optimization suggestions on their long running PR only once it gets stabilized is a genuine ask.
We can achieve this either by having some keyword in commit msg or PR title, or can do it through labels to pause the codeflash untill the PR is stabilizes. Let me know your thoughts?

misrasaurabh1:
i am fine with it but i don't this is the best UI. People would have to think and add labels which does not come naturally. So users will only add this once they get spammed and irritated, which means it is already too late.
After this users will forget to remove this which will disable codeflash permanently for that PR. 
Is there a better way to do this?

KRRT7:
Closing as stale — this PR has been open for 8+ months with no recent activity. Reopen if still relevant.

PR #294, opened 2025-06-06
Add the test case timing to the generated test

misrasaurabh1:
It creates PRs like https://github.com/codeflash-ai/codeflash/pull/293 where the generated tests show all the timings.


misrasaurabh1:
codeflash just sped up vibe code claude code :)

PR #363, opened 2025-06-22
path normalization and tempdir fixes for windows

KRRT7:
normalize paths for windows where necessary


misrasaurabh1:
can you run atleast one e2e and unit test on windows? otherwise we will keep breaking this. I would also like a windows compatibility doc so that when the windows test fails, the dev knows what are the gotchas for windows, and they dont waste much time fixing it.

KRRT7:
CF-427

PR #375, opened 2025-06-24
cleanup properly

misrasaurabh1:
also add verbose mode to --verify-setup

### DIFF
--- a/codeflash/cli_cmds/cmd_init.py
+++ b/codeflash/cli_cmds/cmd_init.py
@@ -34,7 +34,7 @@ if TYPE_CHECKING:
     from argparse import Namespace
 
 CODEFLASH_LOGO: str = (
-    f"{LF}"  # noqa: ISC003
+    f"{LF}"
     r"                   _          ___  _               _     " + f"{LF}"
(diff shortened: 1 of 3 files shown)


codeflash/cli_cmds/cmd_init.py
```diff
@@ -34,7 +34,7 @@
     from argparse import Namespace
 
 CODEFLASH_LOGO: str = (
-    f"{LF}"  # noqa: ISC003
+    f"{LF}"
```
misrasaurabh1:
```suggestion
    f"{LF}" # noqa: ISC003
```

codeflash/cli_cmds/cmd_init.py
```diff
@@ -34,7 +34,7 @@
     from argparse import Namespace
 
 CODEFLASH_LOGO: str = (
-    f"{LF}"  # noqa: ISC003
+    f"{LF}" # noqa: ISC003
```
misrasaurabh1:
```suggestion
    f"{LF}"  # noqa: ISC003
```

PR #488, opened 2025-07-02
 runtime timing info even if `codeflash_output` assignment is missing

aseembits93:


misrasaurabh1 (approved):

aseembits93:
have a newer one with more bugfixes

PR #572, opened 2025-07-23
[FIX] Get fork property in github event correctly

mohammedahmed18:
## How I tested it:

created a fork repo and raised a pull request in the base repo, wrote a github action to save the events.json file as an artifact

image

for some odd reason pull_request.fork is giving false (it was working before maybe they changed it), the pull_request.head.repo.fork though is returning true, so we can use that.

image

### DIFF
--- a/codeflash/code_utils/env_utils.py
+++ b/codeflash/code_utils/env_utils.py
@@ -111,7 +111,7 @@ def get_cached_gh_event_data() -> dict[str, Any]:
 
 def is_repo_a_fork() -> bool:
     event = get_cached_gh_event_data()
-    return bool(event.get("repository", {}).get("fork", False))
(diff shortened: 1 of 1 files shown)


Saga4 (approved):

codeflash/code_utils/env_utils.py
```diff
@@ -111,7 +111,7 @@ def get_cached_gh_event_data() -> dict[str, Any]:
 
 def is_repo_a_fork() -> bool:
     event = get_cached_gh_event_data()
-    return bool(event.get("repository", {}).get("fork", False))
```
Saga4, line 114: I think this must be to check the Repository on which the PR is created to (destination) and not from the source.

mohammedahmed18: yeah looks like this, although this was actually giving true before for forked repository, maybe it was a bug on their end

PR #620, opened 2025-08-06
Pre commit update

KRRT7:
### **PR Type**
Enhancement

___

### **Description**
- Refresh lint config: bump ruff to v0.12.7, add UP045 ignore

- Style fixes: added `return None`, starred tuple unpack, removed unused import

- Refine critic signature: make parameter keyword-only default False

- Add noqa to TestResults class, bump version placeholder

___

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody></tr></tbody></table>

</details>

___

### DIFF
--- a/codeflash/result/critic.py
+++ b/codeflash/result/critic.py
@@ -40,10 +40,8 @@ def speedup_critic(
     The noise floor is doubled when benchmarking on a (noisy) GitHub Action virtual instance, also we want to be more confident there.
     """
     noise_floor = 3 * MIN_IMPROVEMENT_THRESHOLD if original_code_runtime < 10000 else MIN_IMPROVEMENT_THRESHOLD
-    if not disable_gh_action_noise:
-        in_github_actions_mode = bool(env_utils.get_pr_number())
-        if in_github_actions_mode:
-            noise_floor = noise_floor * 2  # Increase the noise floor in GitHub Actions mode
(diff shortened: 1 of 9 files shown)


codeflash/version.py
```diff
@@ -1,2 +1,2 @@
 # These version placeholders will be replaced by uv-dynamic-versioning during build.
-__version__ = "0.16.1"
```
misrasaurabh1: we should not add this change

KRRT7: oops, reverted

codeflash/result/critic.py
```diff
@@ -29,7 +29,8 @@ def speedup_critic(
     candidate_result: OptimizedCandidateResult,
     original_code_runtime: int,
     best_runtime_until_now: int | None,
-    disable_gh_action_noise: Optional[bool] = None,
+    *,
```
aseembits93, line 32: the reason for the '*'?

KRRT7: https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/
and 
https://docs.astral.sh/ruff/rules/boolean-default-value-positional-argument/

aseembits93:
@KRRT7 tests failing

PR #631, opened 2025-08-07
Warning Message for users using old versions

mihikap01:
Prints warning messages when the user tries to run an old version of codeflash.


codeflash/version.py
```diff
@@ -1,2 +1,6 @@
 # These version placeholders will be replaced by uv-dynamic-versioning during build.
+<<<<<<< HEAD
```
Saga4: we can remove this file from local version testing.

Saga4: @mihikap01 we can remove this file from the scope of this PR.

codeflash/code_utils/version_check.py
```diff
@@ -0,0 +1,96 @@
...
+    suggesting the user upgrade.
+    
+    Args:
+        disable_check: If True, skip the version check entirely.
+    """
+    if disable_check:
+        return
+        
+    # Get current version dynamically to handle runtime changes
+    from codeflash.version import __version__ as current_version
```
Saga4: we already have this as global import.

codeflash/main.py
```diff
@@ -21,6 +22,17 @@ def main() -> None:
         CODEFLASH_LOGO, panel_args={"title": "https://codeflash.ai", "expand": False}, text_args={"style": "bold gold3"}
     )
     args = parse_args()
+    
+    # Check for newer version (skip for version command to avoid confusion)
+    if not args.version:
```
Saga4, line 27: With version check 'codeflash --version' we can also show the status if its outdated.

codeflash/main.py
```diff
@@ -21,6 +22,17 @@ def main() -> None:
...
+        # Check if version check is disabled in config
+        disable_version_check = False
```
Saga4, line 29: I am not sure we are going to add this in the configs or existing configs with customers.
We can just have a direct non blocking check.

codeflash/code_utils/version_check.py
```diff
@@ -0,0 +1,96 @@
...
+    global _version_cache
+    
+    # Check cache first
+    current_time = time.time()
+    if (_version_cache["version"] is not None and 
+        current_time - _version_cache["timestamp"] < _cache_duration):
+        return _version_cache["version"]
+    
+    try:
+        response = requests.get("https://pypi.org/pypi/codeflash/json", timeout=10)
```
Saga4: nit: can have quick timeout here to make it non blocking if there is network issue.

Saga4:
Tested here:
```
(cli) saga4@codeflashs-MacBook-Air cli % python -c "from codeflash.code_utils.version_check import check_for_newer_minor_version, get_latest_version_from_pypi
print('Current version:', '0.16.3')  # from version.py
latest = get_latest_version_from_pypi()
print('Latest version from PyPI:', latest)"

Current version: 0.16.3
Latest version from PyPI: 0.16.3
```

PR #649, opened 2025-08-11
[FEAT][LSP] Worktrees

mohammedahmed18:
- **getting worktree to work in both lsp and normal cli mode**
- **worktrees changes**

### DIFF
--- a/codeflash/code_utils/git_utils.py
+++ b/codeflash/code_utils/git_utils.py
@@ -9,13 +9,14 @@ import time
 from functools import cache
 from io import StringIO
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 import git
 from rich.prompt import Confirm
 from unidiff import PatchSet
 
 from codeflash.cli_cmds.console import logger
+from codeflash.code_utils.compat import codeflash_cache_dir
 from codeflash.code_utils.config_consts import N_CANDIDATES
 
 if TYPE_CHECKING:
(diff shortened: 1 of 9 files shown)


codeflash/code_utils/git_utils.py
```diff
@@ -192,3 +194,88 @@ def get_last_commit_author_if_pr_exists(repo: Repo | None = None) -> str | None:
...
+    result = subprocess.run(
+        ["git", "worktree", "add", "-d", str(worktree_dir)],
+        cwd=git_root,
+        check=True,
+        stdout=subprocess.DEVNULL if is_LSP_enabled() else None,
+        stderr=subprocess.DEVNULL if is_LSP_enabled() else None,
+    )
+    if result.returncode != 0:
+        logger.error(f"Failed to create worktree: {result.stderr}")
+        return None
```
KRRT7: can't we do this via the git module we're already using?

codeflash/code_utils/git_utils.py
```diff
@@ -192,3 +194,88 @@ def get_last_commit_author_if_pr_exists(repo: Repo | None = None) -> str | None:
...
+    if not uni_diff_text.strip():
+        logger.info("No uncommitted changes to copy to worktree.")
+        return worktree_dir
+
+    # Write the diff to a temporary file
+    with tempfile.NamedTemporaryFile(mode="w+", suffix=".codeflash.patch", delete=False) as tmp_patch_file:
+        tmp_patch_file.write(uni_diff_text + "\n")  # the new line here is a must otherwise the last hunk won't be valid
+        tmp_patch_file.flush()
+
+        patch_path = Path(tmp_patch_file.name).resolve()
```
KRRT7: since we're using a namedtempfile, we should use `w` mode directly since the file wouldn't exist and thus not truncated

codeflash/models/models.py
```diff
@@ -349,6 +349,9 @@ class OptimizedCandidate:
     explanation: str
     optimization_id: str
 
+    def set_explanation(self, new_explanation: str) -> None:
+        object.__setattr__(self, "explanation", new_explanation)
+
```
KRRT7: prefer [setattr](https://docs.python.org/3/library/functions.html#setattr)

mohammedahmed18: I actually removed the frozen attribute from `OptimizedCandidate`, so there is no need for this hacky way of mutating the explanation

KRRT7 (dismissed):
approving, we should aim to have more test coverage for the LSP once we have it a bit more polished

mohammedahmed18:
@KRRT7 okay let me just fix this mypy issue

KRRT7 (approved):

PR #657, opened 2025-08-14
extend the exclusions, not replace.

KRRT7:
a vanilla `extend` replaces the built in ruff https://docs.astral.sh/ruff/settings/#exclude exclusion rules, we want to extend instead.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -200,7 +200,7 @@ target-version = "py39"
 line-length = 120
 fix = true
 show-fixes = true
-exclude = ["code_to_optimize/", "pie_test_set/", "tests/"]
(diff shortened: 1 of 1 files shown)


misrasaurabh1 (approved):

aseembits93:
@KRRT7 what is the consequence of this change?

KRRT7:
> @KRRT7 what is the consequence of this change?

generally shouldn't notice it. with the current configuration ruff was linting / formatting files in my `.git` folder (which ruff ignores by default but we were replacing it with our own list,  you can see what ruff ignores by default  at https://docs.astral.sh/ruff/settings/#exclude)

PR #661, opened 2025-08-15
Abstracted Clean candidate loop using queue.Queue()

aseembits93:


codeflash/optimization/function_optimizer.py
```diff
@@ -401,48 +477,23 @@ def determine_best_candidate(
...
+            if candidate is None:
+                logger.debug("everything done, exiting")
+                break
+
+            try:
                 candidate_index += 1
                 get_run_tmp_file(Path(f"test_return_values_{candidate_index}.bin")).unlink(missing_ok=True)
                 get_run_tmp_file(Path(f"test_return_values_{candidate_index}.sqlite")).unlink(missing_ok=True)
-                logger.info(f"Optimization candidate {candidate_index}/{original_len}:")
+                logger.info(f"Optimization candidate {candidate_index}/{processor.original_len}:")
```
misrasaurabh1: why original_len? shouldn't it be the length of all candidates?

aseembits93: updated variable name

PR #683, opened 2025-08-23
[FIX] Prevent applying global assignments again when reverting the helpers

mohammedahmed18:
Since we revert helpers by applying the cached original file code back to the file, any existing global assignments in the cached code would get reapplied like here  https://github.com/mohammedahmed18/unstructured/pull/1/files.

The fix was to skip adding global assignments during the revert process.

I don't think this is the best way for reverting helpers, so this is a temporary fix, until we have a better way for doing so


tests/test_code_replacement.py
```diff
@@ -1783,7 +1782,6 @@ def new_function2(value):
 """
     expected_code = """import numpy as np
 
-print("Hello world")
```
mohammedahmed18, line 1786: we are only adding the unique global statements, since the exact print statement was in both original and optimized code, we should get only one statement in the final code not two

tests/test_code_replacement.py
```diff
@@ -3453,3 +3447,157 @@ def hydrate_input_text_actions_with_field_names(
     main_file.unlink(missing_ok=True)
 
     assert new_code == expected
+
+def test_duplicate_global_assignments_when_reverting_helpers():
```
mohammedahmed18, line 3451: the unstructured bug: https://github.com/mohammedahmed18/unstructured/pull/1.

PR #699, opened 2025-08-30
[INIT] install VSCode extension

mohammedahmed18:


codeflash/cli_cmds/cmd_init.py
```diff
@@ -802,6 +805,37 @@ def install_github_actions(override_formatter_check: bool = False) -> None:  # n
...
+                timeout=60,
+                capture_output=True,
+            )
+        except subprocess.TimeoutExpired:
+            error = "Installation timed out."
+        except subprocess.CalledProcessError as e:
+            error = e.stderr or "Unknown error."
+
+    if error:
+        click.echo("❌ Failed to install Codeflash for VSCode.")
```
Saga4: Rather than failed msg, we can suggest user  to install VSC extension and error msg can go to our telemetry for our own logs.

mohammedahmed18: done :+1:

codeflash/cli_cmds/cmd_init.py
```diff
@@ -802,6 +805,41 @@ def install_github_actions(override_formatter_check: bool = False) -> None:  # n
         apologize_and_exit()
 
 
+def install_vscode_extension() -> None:
+    vscode_path = shutil.which("code")
```
misrasaurabh1: does vscode automatically install the "code" cli or does the user have to install it manually by following this step?
my question is that is it justified to always assume that `code` exists? If not , is there a more robust way to discover the installation?

codeflash/cli_cmds/cmd_init.py
```diff
@@ -802,6 +805,41 @@ def install_github_actions(override_formatter_check: bool = False) -> None:  # n
...
+        return
+
+    error = ""
+    with progress_bar("Installing Codeflash for VSCode…"):
+        try:
+            result = subprocess.run(
+                [vscode_path, "--install-extension", "codeflash.codeflash", "--force"],
+                check=True,
+                text=True,
+                timeout=60,
```
misrasaurabh1: if it really takes that much time, i am also thinking that we can run it in parallel with the other init steps and only print that `successfully installed the vs code extension`.
I am concerned about the blocking experience here

codeflash/cli_cmds/cmd_init.py
```diff
@@ -802,6 +805,41 @@ def install_github_actions(override_formatter_check: bool = False) -> None:  # n
         apologize_and_exit()
 
 
+def install_vscode_extension() -> None:
+    vscode_path = shutil.which("code")
+    if not vscode_path:
+        return
+
+    error = ""
+    with progress_bar("Installing Codeflash for VSCode…"):
```
misrasaurabh1: one more thing - what if they have cursor installed? we should also support installing codeflash there

codeflash/cli_cmds/cmd_init.py
mohammedahmed18: on linux systems, if you use your package manager like apt or pacman ...
you will get the cli in your path without doing extra step.

for macos users they need to add it to their path manually after the installation
https://code.visualstudio.com/docs/setup/mac#_configure-the-path-with-vs-code


another workaround, without using the cli is to manually download the vsix file, unzip it, place it in the vscode installation path, in linux it would be:

- ~/.vscode/extensions (official vscode)
- ~/.vscode-oss/extensions (codium)
- ~/.windsurf/extensions
- ~/.cursor/extensions
(not sure about macos but it should be in Application Support dir)

then manually add the codeflash identifier to the extensions.json file.

mohammedahmed18: but I guess we would need this workaround eventually specially for cursor: https://forum.cursor.com/t/command-line-list-extensions/103565/13

so I'll add this manual installation as a fallback

codeflash/cli_cmds/cmd_init.py
mohammedahmed18: it doesn't the extension is around 250kb

mohammedahmed18: it takes 2 seconds to initialize the cli and install the extension

codeflash/cli_cmds/cmd_init.py
mohammedahmed18: yeah, I think for cursor we would need to manually add it because the cli is not working for many people (including me) https://github.com/codeflash-ai/codeflash/pull/699#discussion_r2316100362

codeflash/cli_cmds/cmd_init.py
mohammedahmed18: but yeah let's run it in parallel, if we are going to manually download and and install the extension https://github.com/codeflash-ai/codeflash/pull/699#discussion_r2316057451

codeflash/cli_cmds/cmd_init.py
mohammedahmed18: but we still need to download the vsix file from somewhere, and it's a bit tricky to do that from the official marketplace, that's why we need to push this to the open vsx registery @Saga4 as they have an api for that

KRRT7: > another workaround, without using the cli is to manually download the vsix file, unzip it, place it in the vscode installation path, in linux it would be:

as someone who has continuously had issues with the code cli over the years on basically all platforms, if we can place the VSIX archive where it belongs directly, and it work out of the box, I much prefer to do that.

misrasaurabh1:
@mohammedahmed18 whats the status with this?

mohammedahmed18:
@misrasaurabh1 going to manually download the extension in .vscode dir as a fallback  if the cli not found

Will start on thet tomorrow

misrasaurabh1:
@Saga4 to review

misrasaurabh1:
@mohammedahmed18 whats going on here?

mohammedahmed18:
@misrasaurabh1 after the windows support I still need to make sure this still works on windows,
will borrow a windows machine and test it

misrasaurabh1:
@mohammedahmed18 if this works for mac then lets merge it, unless you have a quick windows machine. After that you can ensure that this works for windows. We should not let PRs get stale.

KRRT7:
let's just merge it, I'll test it on windows later and merge any necessary fixes.

mohammedahmed18:
@KRRT7 okay it should be ready then

PR #718, opened 2025-09-05
[LSP] stderr verbose logs for the thought process

mohammedahmed18:


codeflash/api/aiservice.py
```diff
@@ -133,7 +133,7 @@ def optimize_python_code(  # noqa: D417
             "repo_name": git_repo_name,
         }
 
-        logger.info("Generating optimized candidates…")
+        logger.info("!lsp|tags|Generating optimized candidates…")
```
Saga4: Maybe we can just have it broken in json object with "message", component. That would be easy for anyone to use.
or lets just drop tags as its confusing.

mohammedahmed18: @Saga4 the idea of tags is to use the same message in console and lsp,
sending a json, we would re-write all logger lines, so I think this is the alternative option but dropping it will just not give us control over displaying the messages in the extension (the markdown)

Saga4: Okay, lets then we can remove "tags" as delimeter as thats not helping, pipe is enough I guess.

codeflash/api/aiservice.py
```diff
@@ -144,10 +144,10 @@ def optimize_python_code(  # noqa: D417
 
         if response.status_code == 200:
             optimizations_json = response.json()["optimizations"]
-            logger.info(f"Generated {len(optimizations_json)} candidate optimizations.")
+            logger.info(f"!lsp|tags|Generated {len(optimizations_json)} candidate optimizations.")
             console.rule()
             end_time = time.perf_counter()
-            logger.debug(f"Generating optimizations took {end_time - start_time:.2f} seconds.")
+            logger.debug(f"!lsp|tags|Generating optimizations took {end_time - start_time:.2f} seconds.")
```
Saga4: I think the messaging could be "Generating possible optimizations took {end_time - start_time:.2f} seconds." 

or "Candidate generation for optimizations took {end_time - start_time:.2f} seconds."

codeflash/api/aiservice.py
mohammedahmed18: yeah right, that would be cleaner

Saga4 (approved):

PR #723, opened 2025-09-09
[FIX] set the project root to the pyproject.toml dir

mohammedahmed18:


misrasaurabh1:
> assume that the pyproject.toml file defines the project root.

This assumption is wrong

aseembits93:
LGTM

PR #726, opened 2025-09-10
filelock should be installed when not using `uv sync`

aseembits93:


KRRT7:
can you refresh the lock file while you're at it? `uv sync -U` 

uv.lock is cross platform btw: https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile

KRRT7 (approved):

PR #739, opened 2025-09-17
Get throughput from output  for async functions

KRRT7:


.github/workflows/pre-commit.yaml
```diff
@@ -1,19 +0,0 @@
-name: Lint
```
misrasaurabh1, line 1: why are we removing this?

codeflash/code_utils/codeflash_wrap_decorator.py
```diff
@@ -36,185 +30,27 @@ def get_run_tmp_file(file_path: Path) -> Path:  # moved from codeflash/code_util
...
 def codeflash_behavior_async(func: F) -> F:
     @wraps(func)
     async def async_wrapper(*args: Any, **kwargs: Any) -> Any:  # noqa: ANN401
         loop = asyncio.get_running_loop()
         function_name = func.__name__
-        line_id = f"{func.__name__}_{func.__code__.co_firstlineno}"
+        line_id = os.environ["CODEFLASH_CURRENT_LINE_ID"]
         loop_index = int(os.environ["CODEFLASH_LOOP_INDEX"])
-        test_module_name, test_class_name, test_name = extract_test_context_from_frame()
+        test_module_name, test_class_name, test_name = extract_test_context_from_env()
```
misrasaurabh1, line 53: in a line below why do we have a default of `0` - `iteration = os.environ.get("CODEFLASH_TEST_ITERATION", "0")`

codeflash/verification/test_runner.py
```diff
@@ -98,7 +98,7 @@ def run_behavioral_tests(
                 coverage_cmd.extend(shlex.split(pytest_cmd, posix=IS_POSIX)[1:])
 
             blocklist_args = [f"-p no:{plugin}" for plugin in BEHAVIORAL_BLOCKLISTED_PLUGINS if plugin != "cov"]
-
+            logger.info(f"{' '.join(coverage_cmd + common_pytest_args + blocklist_args + result_args + test_files)}")
```
misrasaurabh1, line 101: info level looks too high to me for this log

codeflash/result/critic.py
```diff
@@ -25,31 +26,73 @@ def performance_gain(*, original_runtime_ns: int, optimized_runtime_ns: int) ->
...
+
+        throughput_is_best = (
+            best_throughput_until_now is None or candidate_result.async_throughput > best_throughput_until_now
+        )
+
+    if original_async_throughput is not None and candidate_result.async_throughput is not None:
+        # When throughput data is available, accept if EITHER throughput OR runtime improves significantly
+        throughput_acceptance = throughput_improved and throughput_is_best
+        runtime_acceptance = runtime_improved and runtime_is_best
+        return throughput_acceptance or runtime_acceptance
```
misrasaurabh1, line 94: i think this might have cases where one improves and the other degrades. This might require more thought in the future

.github/workflows/pre-commit.yaml
KRRT7: I'll restore it later. it's behaving differently locally vs CI

codeflash/result/critic.py
KRRT7: yes, currently, runtime "degrades" https://github.com/codeflash-ai/optimize-me/pull/121 but throughput improves, this still needs work.

PR #758, opened 2025-09-24
[Chore] Async implementation for perform_function_optimization

HeshamHM28:


codeflash/lsp/beta.py
```diff
@@ -325,154 +320,15 @@ def on_patch_applied(_server: CodeflashLanguageServer, params: OnPatchAppliedPar
...
-        if not is_successful(initialization_result):
-            return {"functionName": params.functionName, "status": "error", "message": initialization_result.failure()}
-
-        should_run_experiment, code_context, original_helper_code = initialization_result.unwrap()
-
-        code_print(
-            code_context.read_writable_code.flat,
-            file_name=current_function.file_path,
-            function_name=current_function.function_name,
-        )
```
mohammedahmed18, line 356: @HeshamHM28 looks like you accidentally removed the code print

codeflash/lsp/beta.py
```diff
@@ -325,154 +320,15 @@ def on_patch_applied(_server: CodeflashLanguageServer, params: OnPatchAppliedPar
...
-            "extra": f"Speedup: {speedup:.2f}x faster",
-            "patch_file": metadata["patch_path"],
-            "patch_id": metadata["id"],
-            "explanation": best_optimization.explanation_v2,
-        }
+        result = await loop.run_in_executor(None, sync_perform_optimization, server, params)
+    except asyncio.CancelledError:
+        return {"status": "canceled", "message": "Task was canceled"}
+    else:
+        return result
```
KRRT7, line 349: this doesn't make sense to me

@server.thread()

& loop.run_in_executor(None, sync_perform_optimization are basically equivalent.

codeflash/lsp/beta.py
```diff
@@ -325,154 +320,15 @@ def on_patch_applied(_server: CodeflashLanguageServer, params: OnPatchAppliedPar
...
-            "status": "success",
-            "message": "Optimization completed successfully",
-            "extra": f"Speedup: {speedup:.2f}x faster",
-            "patch_file": metadata["patch_path"],
-            "patch_id": metadata["id"],
-            "explanation": best_optimization.explanation_v2,
-        }
+        result = await loop.run_in_executor(None, sync_perform_optimization, server, params)
+    except asyncio.CancelledError:
+        return {"status": "canceled", "message": "Task was canceled"}
```
KRRT7, line 347: for this to work, there would need to be an event that raises the cancellation event, there's nothing currently raising it.

I think a potential solution would be to use a threading.Event.

codeflash/lsp/features/perform_optimization.py
```diff
@@ -0,0 +1,152 @@
+import contextlib
```
KRRT7, line 1: I think this extra layer of indirection isn't necessary, there should be a cleaner impl

codeflash/lsp/beta.py
HeshamHM28: With `server.thread()`, you cannot kill the thread. Based on the PyGLS implementation, if the function is iscoroutine, it’s stored in a dictionary, and when we send a cancellation, PyGLS stops it. @KRRT7

codeflash/lsp/beta.py
HeshamHM28: The event is raised in PyGLS itself. When the client cancels the request, it raises asyncio.CancelledError

KRRT7: if PyGLS can raise an cancellation mechanism for asynchronous tasks, then it should be available for synchronous tasks as well

codeflash/lsp/beta.py
mohammedahmed18: @KRRT7 
the thread() is using ThreadPoolExecutor in which you can't cancel the task if it started executing

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.cancel

codeflash/lsp/beta.py
mohammedahmed18: @KRRT7 
the only cancellable tasks are the ones added to the _request_futures dictionary, take a look:
https://github.com/openlawlibrary/pygls/blob/805ae6f57494af75b8934745af47141817149cc6/pygls/protocol/json_rpc.py#L176

codeflash/lsp/beta.py
KRRT7: @mohammedahmed18 yes, and loop.run_in_executor simply creates a concurrent.futures.ThreadPoolExecutor, if it doesn't already exists: https://github.com/python/cpython/blob/3.13/Lib/asyncio/base_events.py#L895-L899

that's why I mentioned earlier that they're equivalent.

mohammedahmed18: right, they are the same, the difference is that the feature function is now a coroutine which -according to pygls implementation- will be transformed into asyncio.Future rather than a [Future](https://github.com/python/cpython/blob/773a7c05ff0c2cc9076140d9cd767e94f7f29018/Lib/concurrent/futures/_base.py#L325) (if it's threaded function)


```py
 # with async
 if asyncio.iscoroutinefunction(handler):
    future = asyncio.ensure_future(handler(*args, **kwargs))
    self._request_futures[msg_id] = future
    future.add_done_callback(callback)

# with thread decorator
elif is_thread_function(handler):
    future = self._server.thread_pool.submit(handler, *args, **kwargs)
    self._request_futures[msg_id] = future
    future.add_done_callback(callback)
```

and the thread() future can't be canceled if it's running: https://github.com/python/cpython/blob/773a7c05ff0c2cc9076140d9cd767e94f7f29018/Lib/concurrent/futures/_base.py#L368

while asyncio.Future can

KRRT7: hadn't noticed! thanks!

codeflash/lsp/beta.py
KRRT7: sounds good, I'll trust you then, the only criticism I have now is the extra indirection

HeshamHM28: I think it’s not really an indirection call. Also, that keeps beta.py clean. We’ll also want to introduce lazy loading, which will help us. What do you think?

KRRT7: i'm working on some lazy loading improvements, i think it'll be more of a structural change so everything will benefit including the LSP<br><br>if you want to ship this in the current state and improve it later on, go for it

HeshamHM28: Could you approve it ?

KRRT7 (approved):

saga4893:
@HeshamHM28 Are we planning to merge this?

HeshamHM28:
> @HeshamHM28 Are we planning to merge this?

yeah

PR #796, opened 2025-10-07
[Worktree] mirror the path for the args.all

mohammedahmed18:


Saga4 (approved):

KRRT7 (changes requested):
modify one of the E2E tests inplace to use --worktree and use that as a E2E  test for worktrees as well

mohammedahmed18:
@KRRT7 will create a separate pr for this later

PR #860, opened 2025-10-28
Collect the same number of tests for behavorial, benchmarking and lp

aseembits93:
For https://github.com/codeflash-ai/datacompy/pull/13
Before
Screenshot 2025-10-28 at 11 38 53
Now
Screenshot 2025-10-28 at 11 35 04


aseembits93:
you might notice that the generated tests are passing for behavior and failing for benchmark, that's a different issue, I'll create a separate PR for it on the backend. @misrasaurabh1

PR #935, opened 2025-11-21
[FIX] Extract the required module names inside the read-writable context (CF-862)

mohammedahmed18:
this resolves two main issues:

1- the global definitions are not extracted inside the read-writable code context resulting frequently into mock variables when generating the tests which led to a failed optimization

2- when deciding for each function dependencies we conflict the class attributes with module-level variables if they have the same name, for example

```py
x = "some value"
class MY_CLASS
    def __init__(self):
        self.x = 1
```

it will say that MY_CLASS depends on the global definition `x` which is not correct

### DIFF
--- a/codeflash/discovery/discover_unit_tests.py
+++ b/codeflash/discovery/discover_unit_tests.py
@@ -868,6 +868,58 @@ def process_test_files(
                     continue
                 try:
                     if not definition or definition[0].type != "function":
+                        # Fallback: Try to match against functions_to_optimize when Jedi can't resolve
+                        # This handles cases where Jedi fails with pytest fixtures
+                        if functions_to_optimize and name.name:
+                            for func_to_opt in functions_to_optimize:
+                                # Check if this unresolved name matches a function we're looking for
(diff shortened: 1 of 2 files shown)


tests/test_code_context_extractor.py
```diff
@@ -1026,11 +1044,98 @@ def helper_method(self):
         ending_line=None,
     )
 
-    # In this scenario, the testgen code context is too long, so we abort.
-    with pytest.raises(ValueError, match="Testgen code context has exceeded token limit, cannot proceed"):
+    # In this scenario, the read-writable code context is too long because the __init_ function is reftencing the global x variable not the class attribute (x), so we abort.
```
misrasaurabh1: typo

aseembits93:
@mohammedahmed18 wait where did you get the codeflash optimization?

mohammedahmed18:
@aseembits93 from the extension by the git commit hook

KRRT7:
@mohammedahmed18 maybe we should add some identifier so we know when the optimizations are done via the VSC extension?

PR #1003, opened 2025-12-31
[FEAT][EXP] Adaptive optimizations (CF-831)

mohammedahmed18:
depends on https://github.com/codeflash-ai/codeflash-internal/pull/2157

### DIFF
--- a/codeflash/verification/comparator.py
+++ b/codeflash/verification/comparator.py
@@ -227,6 +227,19 @@ def comparator(orig: Any, new: Any, superset_obj=False) -> bool:  # noqa: ANN001
             if isinstance(orig, np.dtype):
                 return orig == new
 
+            # Handle numpy random generators
+            if isinstance(orig, np.random.Generator):
+                # Compare the underlying BitGenerator state
+                orig_state = orig.bit_generator.state
+                new_state = new.bit_generator.state
+                return comparator(orig_state, new_state, superset_obj)
+
+            if isinstance(orig, np.random.RandomState):
+                # Compare the internal state
+                orig_state = orig.get_state(legacy=False)
+                new_state = new.get_state(legacy=False)
+                return comparator(orig_state, new_state, superset_obj)
+
         if HAS_SCIPY and isinstance(orig, scipy.sparse.spmatrix):
             if orig.dtype != new.dtype:
                 return False
--- a/tests/test_comparator.py
+++ b/tests/test_comparator.py
@@ -360,6 +360,81 @@ def test_numpy():
     assert not comparator(a_void, c_void)
 
 
+def test_numpy_random_generator():
+    try:
+        import numpy as np
+    except ImportError:
+        pytest.skip()
+
+    # Test numpy.random.Generator (modern API)
+    # Same seed should produce equal generators
+    rng1 = np.random.default_rng(seed=42)
+    rng2 = np.random.default_rng(seed=42)
+    assert comparator(rng1, rng2)
+
+    # Different seeds should produce non-equal generators
+    rng3 = np.random.default_rng(seed=123)
+    assert not comparator(rng1, rng3)
+
+    # After generating numbers, state changes
+    rng4 = np.random.default_rng(seed=42)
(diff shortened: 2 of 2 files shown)


codeflash/optimization/function_optimizer.py
```diff
@@ -1004,6 +1081,52 @@ def determine_best_candidate(
...
+        if self.adaptive_optimization_counter >= MAX_ADAPTIVE_OPTIMIZATIONS_PER_TRACE:
+            logger.debug(
+                f"Max adaptive optimizations reached for {self.function_to_optimize.qualified_name}: {self.adaptive_optimization_counter}"
+            )
+            return None
+
+        prev_candidates = candidate_node.path_to_root()
+        if len(prev_candidates) == 1:
+            # we already have the refinement going for this single candidate tree, no need to do adaptive optimize
+            return None
```
mohammedahmed18: TODO: instead of checking the tree length, check if there was a refinement before, and only call the adaptive endpoint if so.

the reason is that calling the adaptive on non-refined candidate is a bit of duplication, as the candidate will be refined anyway

so we can end up having these two processes in parallel

optimize (c1) -> repair (c2) -> adaptive (c3)
optimize (c1) -> repair (c2) -> refine (c3) -> adaptive (c4)

resulting in 3 different candidates (2 adaptives, 1 refinement)

it's better to only have
optimize (c1) -> repair (c2) -> refine (c3) -> adaptive (c4)

which will result in 2 candidates (1 refined , 1 adaptive)

mohammedahmed18: @misrasaurabh1 @aseembits93 @KRRT7 
what do you think guys, when the adaptive should be called

KRRT7 (commented):
## Code Review: Adaptive Optimizations

### Potential Issue: Forest Cycle/Corruption

In `CandidateForest.add()`, if `/adaptive_optimize` returns a candidate with a `parent_id` that doesn't exist yet, a placeholder node is created:

```python
if parent is None:
    parent = CandidateNode(candidate=None)  # placeholder
    self.nodes[pid] = parent
```

This placeholder has `candidate=None`, but `path_to_root()` unconditionally appends `node.candidate`:

```python
def path_to_root(self) -> list[OptimizedCandidate]:
    path = []
    node: CandidateNode | None = self
    while node:
        path.append(node.candidate)  # Will append None for placeholder
        node = node.parent
    return path[::-1]
```

Then in `call_adaptive_optimize()`, the code iterates over this path and accesses `c.optimization_id`, `c.source_code`, etc. - which would raise `AttributeError` if `c` is `None`.

**Question**: What prevents the adaptive optimize endpoint from returning a candidate with a `parent_id` that hasn't been processed yet? If this can happen, `path_to_root()` needs to filter out `None` candidates or the placeholder logic needs rethinking.

KRRT7:
used Claude to help me review here, testing it for that

mohammedahmed18:
@KRRT7 in adaptive optimization the new candidate's parent is the last candidate in the tree
so any adaptive optimization we are sure it has a parent

https://github.com/codeflash-ai/codeflash-internal/blob/5d938dd3f9afe86c9d163eef97b4c28ae2ac7388/django/aiservice/adaptive_optimizer/adaptive_optimizer.py#L103-L105

KRRT7 (approved):

aseembits93:
@mohammedahmed18 why do we need a new endpoint for adaptive optimization? why can't we modify existing endpoints to take past optimizations into consideration?

KRRT7:
@aseem then we'll have to keep those optimizations in memory in the backend which would be tricky to implement i guess

mohammedahmed18:
@aseembits93 @KRRT7 actually I'm using the same optimizer context used in optimize & line-profiler optimize.

https://github.com/codeflash-ai/codeflash-internal/blob/main/django/aiservice/adaptive_optimizer/adaptive_optimizer_context.py

the duplication we have in all 5 endpoints can be refactored but maybe not in a single endpoint imo, this will make it hard for us to implement something specific for a single endpoint.

PR #1031, opened 2026-01-09
feat: add --limit option to codeflash optimize -m pytest

KRRT7:
## Summary
- Adds `--limit=<number>` option to limit how many test files are processed when running `codeflash optimize -m pytest <path>`


codeflash/tracing/pytest_parallelization.py
```diff
@@ -58,6 +58,10 @@ def pytest_split(
     test_files = list(test_files)
     shuffle(test_files)
 
+    # Apply limit if specified
```
aseembits93, line 61: @KRRT7  I wonder if there could be a smarter way of limiting, counting subdirectory wise or in alphabetical order

KRRT7: yeah that's something that could be done in the future, for now I want to keep it as it is for debugging purposes

aseembits93 (approved):

PR #1032, opened 2026-01-09
refresh lockfile

KRRT7:


aseembits93:
@KRRT7 what was the command used to create it, and the python version of the environment? asking for future consistency in updating the lockfile.

KRRT7:
@aseembits93 `uv sync --all-extras --all-groups --upgrade` to upgrade everything at once, however you don't need to specify the python version, uv does that resolution automatically and encodes per-version information into the lock file, so that when you run it against a specific python version resolution is faster, for the ones it doesn't have data on, it does the resolution and adds it to the lock file

PR #1038, opened 2026-01-09
Add numba and tensorflow in test dependency group

aseembits93:


aseembits93:
@KRRT7 shall I upgrade the lock file as well? I use this way of doing it
```
uv venv --python 3.9
uv sync --all-extras --all-groups --all-packages
```

PR #1042, opened 2026-01-12
feat: add --server flag and set local AI service timeout

KRRT7:
## Summary
- Add `--server` CLI flag to switch between local/prod AI service
- Set 5 min timeout for local AI service to prevent indefinite hangs

## Usage
```bash
codeflash --file foo.py --server local
codeflash --file foo.py --server prod
```


codeflash/cli_cmds/cli.py
```diff
@@ -107,6 +108,12 @@ def parse_args() -> Namespace:
         action="store_true",
         help="(Deprecated) Async function optimization is now enabled by default. This flag is ignored.",
     )
+    parser.add_argument(
```
misrasaurabh1, line 111: why this over the env var method?

misrasaurabh1: we don't really want to make this part of the user interface for codeflash cli

codeflash/cli_cmds/cli.py
```diff
@@ -121,6 +128,9 @@ def process_and_validate_cmd_args(args: Namespace) -> Namespace:
     )
     from codeflash.code_utils.github_utils import require_github_app_or_exit
 
+    if args.server:
+        os.environ["CODEFLASH_AIS_SERVER"] = args.server
```
misrasaurabh1, line 135: this can go wrong because there is no validation of user input to be only prod/local

codeflash/cli_cmds/cli.py
KRRT7: A/B testing, i want the flexibility so that I don't have to touch my env vars

codeflash/cli_cmds/cli.py
KRRT7: it's meant to be for internal use only so I think we'll be ok

PR #1043, opened 2026-01-12
upgrade cli to review prompt changes

Saga4:
- **remove unused diff from cli for review**
- **update**


KRRT7:
what's the before and after look like?

Saga4:
> what's the before and after look like?

The output remains same, its just we are providing more context. And changes in aiservice are compatible to older versions too.
Tested this earlier with `codeflash --file src/math/number_theory.py --function gcd_recursive --no-pr`

before:

image


After:
```
## 2. Code Quality Evaluation │ │ │ │ **Optimization Technique**: Recursion → Iteration transformation │ │ │ │ **Readability Assessment**: │ │ - **Original**: Classic textbook recursive Euclidean algorithm (4 lines) │ │ - **Optimized**: Iterative while-loop with tuple unpacking (3 lines) │ │ - **Verdict**: Both versions are **equally clear and idiomatic**. The iterative version is actually the more common implementation in production │ │ code and arguably more readable (no mental stack tracking needed). │ │ │ │ **Maintainability**: │ │ - Algorithm complexity unchanged (still O(log(min(a,b)))) │ │ - Semantics identical (same mathematical operation) │ │ - No side effects introduced │ │ - Code is actually slightly simpler (no recursion depth concerns) │ │ │ │ **Anti-Pattern Check**: │ │ - ✅ No global or nonlocal keywords │ │ - ✅ Not a pure micro-optimization - eliminates fundamental recursion overhead │ │ - ✅ Algorithmic technique change (recursion elimination is a recognized optimization strategy) │ │ - ✅ Not import-only changes │
```

KRRT7 (approved):

PR #1044, opened 2026-01-13
More stringent unit tests for gpu-instrumentation

aseembits93:


aseembits93:
code normalization is needed as `ast.unparse` resolves nested strings with double quotes and/or single quotes differently across different python versions

PR #1051, opened 2026-01-14
AST Parser for detecting numerical code (numpy/scipy/torch/tensorflow/jax/numba)

aseembits93:
In the case of numpy/scipy/math it will return False when numba is not present in the environment because this PR is part of a system which will try to use Just-in-Time compilation which is part of numba to compile numpy/scipy/math functions.

### DIFF
--- a/codeflash/verification/comparator.py
+++ b/codeflash/verification/comparator.py
@@ -296,6 +296,9 @@ def comparator(orig: Any, new: Any, superset_obj=False) -> bool:  # noqa: ANN001
             if isinstance(orig, torch.dtype):
                 return orig == new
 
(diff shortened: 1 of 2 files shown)


aseembits93:
@claude review this

KRRT7:
As with the other PR you can skip anything async for this, these sort of things apply mainly for CPU bound things, which asyncio doesn't do

aseembits93:
> As with the other PR you can skip anything async for this, these sort of things apply mainly for CPU bound things, which asyncio doesn't do

Agreed, I have removed async visitors here. ready to review

mohammedahmed18 (approved):

PR #1053, opened 2026-01-14
Comparator for torch.device

aseembits93:


tests/test_comparator.py
```diff
@@ -824,6 +824,50 @@ def test_torch():
...
+    if torch.cuda.is_available():
+        assert not comparator(c, d)
+
+    # Test device with index
+    e = torch.device("cpu")
+    f = torch.device("cpu")
+    assert comparator(e, f)
+
+    # Test cuda devices with different indices (if multiple GPUs available)
+    if torch.cuda.is_available() and torch.cuda.device_count() > 1:
```
aseembits93, line 850: this will not run in CI, I have a way of mocking it

PR #1054, opened 2026-01-14
display RO / RW context (--verbose)

KRRT7:


KRRT7:
I think token_limit is currently too low, thoughts @aseembits93

codeflash/discovery/discover_unit_tests.py
```diff
@@ -508,7 +508,7 @@ def analyze_imports_in_test_file(test_file_path: Path | str, target_functions: s
         return True
 
     if analyzer.found_any_target_function:
-        logger.debug(f"Test file {test_file_path} imports target function: {analyzer.found_qualified_name}")
+        # logger.debug(f"Test file {test_file_path} imports target function: {analyzer.found_qualified_name}")
```
Saga4, line 511: why to keep the logs as comments when commenting.

KRRT7: just in case we want to bring them back

KRRT7:
Screenshot 2026-01-14 at 8 11 32 AM
example

Saga4:
> I think token_limit is currently too low, thoughts @aseembits93

Did we got quote increase from foundary? 
And when did we see the tokens exceeding like in which scenerio? Test gen etc?

KRRT7:
@Saga4 I noticed functions that call a lot of other functions brings in a lot of context which caused it to hit the cap somewhere

KRRT7:
well --verbose is only used for debugging so far which is where it fits best imo

aseembits93:
looking into token limits right now

codeflash/code_utils/config_consts.py
```diff
@@ -4,6 +4,8 @@
 from typing import Any, Union
 
 MAX_TEST_RUN_ITERATIONS = 5
+OPTIMIZATION_CONTEXT_TOKEN_LIMIT = 16000
+TESTGEN_CONTEXT_TOKEN_LIMIT = 16000
```
aseembits93, line 8: @KRRT7 
```suggestion
OPTIMIZATION_CONTEXT_TOKEN_LIMIT = 50000 # 25% of the context length limit of 200K of modern LLMs
TESTGEN_CONTEXT_TOKEN_LIMIT = 50000 # 25% of the context length limit of 200K of modern LLMs 
```

aseembits93:
@Saga4 @KRRT7 from a cursory look at recent testgen and optimization generation calls in the observability tool, I'm seeing approximately 30k tokens at max in the input. This is far below the ~200k input token thresholds modern LLMs.

PR #1058, opened 2026-01-15
codeflash-omni

misrasaurabh1:


codeflash/optimization/function_optimizer.py
```diff
@@ -2460,7 +2484,17 @@ def run_optimized_candidate(
                 else 0
             )
 
-            if (total_candidate_timing := candidate_benchmarking_results.total_passed_runtime()) == 0:
+            total_candidate_timing = candidate_benchmarking_results.total_passed_runtime()
+
+            # For JavaScript/TypeScript: If performance benchmarking fails, use behavioral test timing as fallback
```
misrasaurabh1: this is a really bad idea

PR #1086, opened 2026-01-16
FIX: --no-gen-tests Flag Causes Path Resolution Error with Module-Root

mashraf-222:
## Fix: Handle missing test file registration gracefully with improved path matching

1. Path normalization: Added a helper that normalizes paths before comparison:
   - Resolves to absolute paths
   - Handles Windows case-insensitivity (only on Windows)
   - Works when files don't exist yet

2. Error handling: Replaced the assertion with a warning that:
   - Logs the missing test file path
   - Lists registered test files for debugging
   - Skips the test case instead of crashing

### Changes
- `models.py`: Added `_normalize_path_for_comparison()` and updated path comparison methods to use it
- `parse_test_output.py`: Replaced assertion with warning and continue

### Impact
- Prevents crashes when test files aren't registered
- Improves debugging with clearer error messages
- Handles Windows path differences correctly
- Minimal code changes, no breaking changes

Fixes the crash and improves robustness when test file paths don't match exactly.


codeflash/models/models.py
```diff
@@ -372,23 +372,58 @@ def add(self, test_file: TestFile) -> None:
...
 
     def get_by_original_file_path(self, file_path: Path) -> TestFile | None:
-        return next((test_file for test_file in self.test_files if test_file.original_file_path == file_path), None)
+        normalized = self._normalize_path_for_comparison(file_path)
+        return next(
+            (
+                test_file
+                for test_file in self.test_files
+                if test_file.original_file_path is not None
+                and normalized == self._normalize_path_for_comparison(test_file.original_file_path)
```
aseembits93: can refactor self._normalize_path_for_comparison(test_file.original_file_path) out for better readability

codeflash/models/models.py
```diff
@@ -372,23 +372,58 @@ def add(self, test_file: TestFile) -> None:
...
+        normalized = self._normalize_path_for_comparison(file_path)
         return next(
             (
                 test_file.test_type
                 for test_file in self.test_files
-                if (file_path in (test_file.instrumented_behavior_file_path, test_file.benchmarking_file_path))
+                if normalized == self._normalize_path_for_comparison(test_file.instrumented_behavior_file_path)
+                or (
+                    test_file.benchmarking_file_path is not None
+                    and normalized == self._normalize_path_for_comparison(test_file.benchmarking_file_path)
```
aseembits93: same feedback as previous comment

codeflash/models/models.py
mashraf-222: Done

codeflash/models/models.py
mashraf-222: Done

aseembits93:
@claude review this

PR #1120, opened 2026-01-20
Comparator for wrapped exceptions

aseembits93:


codeflash/verification/comparator.py
```diff
@@ -26,9 +26,56 @@
 HAS_TENSORFLOW = find_spec("tensorflow") is not None
 
 
+def _get_wrapped_exception(exc: BaseException) -> Optional[BaseException]:  # noqa: FA100
+    """Get the wrapped exception if this is a simple wrapper.
```
KRRT7, line 50: I wonder why your agents aren't picking up 

https://github.com/codeflash-ai/codeflash/blob/main/AGENTS.md?plain=1#L72

KRRT7: maybe I should just rename it to Claude.md since we've all started using it, will do that

PR #1140, opened 2026-01-22
CF-1022 pyinstaller + Nuitka support

KRRT7:


Saga4:
A couple of questions
1. How are you testing the binaries created? 
2. And with npm package, we will publish a script to download these from some server location based on the platform?

KRRT7:
@Saga4 

> How are you testing the binaries created?

locally and linux / windows servers

> And with npm package, we will publish a script to download these from some server location based on the platform?

I'm not sure what that will look like yet, I need to see exactly how omni is doing things before moving on, one step at a time.

PR #1146, opened 2026-01-23
reduce invalid crosshair tests noise

KRRT7:


KRRT7:
merging in now to reduce crosshairs spam

PR #1164, opened 2026-01-23
Disable jit during tracing

aseembits93:


aseembits93:
@KRRT7 claude review should have appeared right?

KRRT7:
should have

PR #1165, opened 2026-01-23
covering bytearray iterator for comparator

aseembits93:


aseembits93:
closing in favor of a more general PR which covers abstract base classes like `Iterable` `Iterator` `Reversible` etc

PR #1201, opened 2026-01-30
[FIX] correctly detect codeflash import

mohammedahmed18:
this fixes the issue where codeflash import is not added because the generated test has the word codeflash

```bash

    ReferenceError: codeflash is not defined

      302 |       }
      303 |
    > 304 |       const result = await codeflash.capture('PlainTextStagingStrategy.save', '8', PlainTextStagingStrategy.save.bind(PlainTextStagingStrategy), context)
          |                      ^
      305 |
      306 |       expect(result).toEqual({ success: true, storageType: 'plain_text' })
      307 |       expect(mockedFindFirst).toHaveBeenCalledTimes(1)

```

### DIFF
--- a/codeflash/languages/javascript/instrument.py
+++ b/codeflash/languages/javascript/instrument.py
@@ -51,6 +51,11 @@ class StandaloneCallMatch:
     has_trailing_semicolon: bool
 
 
+codeflash_import_pattern = re.compile(
+    r"(import\s+codeflash\s+from\s+['\"]codeflash['\"])|(const\s+codeflash\s*=\s*require\(['\"]codeflash['\"]\))"
+)
+
+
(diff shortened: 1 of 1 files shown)


codeflash/languages/javascript/instrument.py
```diff
@@ -51,6 +51,11 @@ class StandaloneCallMatch:
     has_trailing_semicolon: bool
 
 
+codeflash_import_pattern = re.compile(
```
Saga4, line 54: There could be nested import in cjs files too like conditional imports etc.

mohammedahmed18: this regex for capturing our own instrumentation which is a basic import
so we are good

PR #1222, opened 2026-01-31
fix: prevent imported types from being duplicated during code replacement

misrasaurabh1:
## Summary

Fix bug where type definitions (interfaces, types) were being duplicated in the optimized output.

## Root Cause Analysis

**The real issue**: In `get_code_optimization_context_for_language`, `code_context.read_only_context` (containing type definitions) was being **prepended to `target_file_code`**, making it part of the **read-writable code** sent to the AI.

When the AI receives:
- `source_code` = code with type definitions prepended (AI thinks this is code to optimize)
- `dependency_code` = `""` (empty - no read-only context!)

The AI naturally includes the type definitions in its output because they're part of the input code.

### 1. Context Extractor Fix (root cause)
Pass `read_only_context` as `read_only_context_code` (dependency_code) instead of prepending it to the source code:
- `source_code` = only the function to optimize
- `dependency_code` = type definitions (AI sees as context, won't include in output)

### 2. Code Replacer Fix (defense-in-depth)
Also check for imported names when adding new declarations from optimized code:
- If a type is already imported, don't add it as a new declaration

## Verification

Before fix:
```
read_writable_code.markdown: contains "interface TreeNode" (BAD!)
read_only_context_code: "" (empty)
```

After fix:
```
read_writable_code.markdown: contains only the function (GOOD!)
read_only_context_code: contains "interface TreeNode" (GOOD!)
```

## Test plan

- [x] Added test `test_imported_interface_not_added_as_declaration` - verifies imported types are not duplicated in code replacer
- [x] Added test `test_multiple_imported_types_not_duplicated` - verifies multiple imported types are not duplicated
- [x] All 54 JavaScript support tests pass
- [x] All 81 code context extractor tests pass
- [x] All 61 code replacer tests pass

## Related

Fixes the bug in https://github.com/codeflash-ai/appsmith/pull/20 where `TreeNode` interface was duplicated.

### DIFF
--- a/codeflash/context/code_context_extractor.py
+++ b/codeflash/context/code_context_extractor.py
@@ -294,10 +294,9 @@ def get_code_optimization_context_for_language(
         helper_code = "\n\n".join(h.source_code for h in same_file_helpers)
         target_file_code = target_file_code + "\n\n" + helper_code
 
-    # Add global variables (module-level declarations) referenced by the function and helpers
-    # These should be included in read-writable context so AI can modify them if needed
-    if code_context.read_only_context:
-        target_file_code = code_context.read_only_context + "\n\n" + target_file_code
+    # Note: code_context.read_only_context contains type definitions and global variables
+    # These should be passed as read-only context to the AI, not prepended to the target code
+    # If prepended to target code, the AI treats them as code to optimize and includes them in output
 
     # Add imports to target file code
     if imports_code:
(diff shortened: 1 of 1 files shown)


codeflash/code_utils/code_replacer.py
```diff
@@ -659,6 +659,19 @@ def _add_global_declarations_for_language(
...
         existing_names = {decl.name for decl in original_declarations}
 
+        # Also exclude names that are already imported (to avoid duplicating imported types)
+        original_imports = analyzer.find_imports(original_source)
+        for imp in original_imports:
+            # Add default import name
+            if imp.default_import:
+                existing_names.add(imp.default_import)
+            # Add named imports (use alias if present, otherwise use original name)
+            for name, alias in imp.named_imports:
```
Saga4: Nit: For this I think we have to add types info as read only context too

tests/test_languages/test_js_code_replacer.py
```diff
@@ -1893,3 +1893,174 @@ class DataProcessor<T> {
...
+
+        When TreeNode is imported from another file and the optimized code
+        contains the TreeNode interface definition (from read-only context),
+        the replacement should NOT add the interface to the original file.
+        """
+        from codeflash.models.models import CodeStringsMarkdown, CodeString
+
+        # Original source imports TreeNode
+        original_source = """\
+import type { TreeNode } from "./constants";
```
Saga4: Lets also add tests for CJS modules

PR #1223, opened 2026-01-31
pytest-xdist in ci

misrasaurabh1:


Saga4 (approved):
nit: we can limit it to run for 4 cores as generally 4 core machine is allocated for GHA

misrasaurabh1:
hard to make this work, closing

PR #1227, opened 2026-01-31
limit the python version for the postinstall script

KRRT7:
otherwise it defaults to 3.14 which doesn't;t have wheels for line-profiler and tries to build from source

### DIFF
--- a/codeflash/discovery/functions_to_optimize.py
+++ b/codeflash/discovery/functions_to_optimize.py
@@ -41,6 +41,8 @@ if TYPE_CHECKING:
 
     from codeflash.models.models import CodeOptimizationContext
     from codeflash.verification.verification_utils import TestConfig
+import contextlib
+
 from rich.text import Text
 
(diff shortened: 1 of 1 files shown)


packages/codeflash/scripts/postinstall.js
```diff
@@ -115,7 +115,7 @@ function installCodeflash(uvBin) {
   try {
     // Use uv tool install to install codeflash in an isolated environment
     // This avoids conflicts with any existing Python environments
-    execSync(`"${uvBin}" tool install codeflash --force`, {
+    execSync(`"${uvBin}" tool install --force --python python3.12 codeflash`, {
```
misrasaurabh1, line 118: not 3.13?

KRRT7: aider has had it in 3.12 for a while, probably just to be safe

KRRT7:
@claude run pre-commit and push the changes

PR #1228, opened 2026-01-31
fix(discovery): use pattern-based filtering when tests_root overlaps source

Saga4:
## Summary

- Fixed critical bug where all functions were incorrectly filtered as "test functions" when `tests_root` equals `module_root` or `project_root`
- Introduced pattern-based test file detection that checks filename patterns (`.test.`, `.spec.`, `_test.`, `_spec.`) and directory patterns (`/test/`, `/tests/`, `/__tests__/`)
- Directory patterns are only checked within the project root to avoid false positives from parent directories (e.g., when repo is located in `/path/tests/my-project`)

## Problem

When auto-detection sets `tests_root` to the same value as `module_root` or `project_root` (common in monorepos and projects without dedicated test directories), the filter condition `file_path.startswith(tests_root + os.sep)` matches ALL source files, causing them all to be filtered out as "test functions".

## Solution

The fix adds an `is_test_file()` helper function that:
1. Detects when tests_root overlaps with source directories
2. Uses pattern-based matching in this case instead of directory-based matching
3. Falls back to the original behavior when tests are in a dedicated separate directory

## Test Plan

- [x] Verified that n8n repo now discovers 1+ optimizable functions instead of 0
- [ ] Run existing test suite to ensure no regressions

### DIFF
--- a/tests/test_function_discovery.py
+++ b/tests/test_function_discovery.py
@@ -604,3 +604,349 @@ def test_function_in_tests_dir():
             assert "vanilla_function" not in remaining_functions
         files_and_funcs = get_all_files_and_functions(module_root_path=temp_dir, ignore_paths=[])
         assert len(files_and_funcs) == 6
+
+
+def test_filter_functions_tests_root_overlaps_source():
+    """Test that source files are not filtered when tests_root equals module_root or project_root.
+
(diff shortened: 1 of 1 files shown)


misrasaurabh1:
Please add tests

PR #1238, opened 2026-02-01
Pre check on env config dependencies in JS before calling aiservices API

Saga4:


misrasaurabh1:
should this also happen during init time? also what happens during init when the deps are not present.

Saga4:
> should this also happen during init time? also what happens during init when the deps are not present.

Will add to the init process too. 
@misrasaurabh1 We have added a new functionality to auto configure codeflash, we can also test it with --show-config 
And also it asks for validation on first run (non Github action) to configure and save it to respective files (package.json or pyproject.toml)

PR #1258, opened 2026-02-02
read xml and cleanup for vitest behavorial test results

Saga4:


Saga4:
Merging this to fix issue with capturePerf for vitest.

PR #1259, opened 2026-02-02
vitest perf looping inefficiency fix

Saga4:
- **read xml and cleanup for vitest perf results**
- **vitest perf looping inefficiency**


Saga4:
Testing on Nextjs project with vitest so require this fix.

PR #1312, opened 2026-02-03
fix: add start tag output to capturePerf for throughput calculation

Saga4:
## Summary
- Adds start tag output (`!$######...######$!`) to capturePerf before timing measurement
- Required for throughput calculation and timing marker matching

## Problem
`capturePerf` was only outputting end tags with timing data (`!######...######!`), but not start tags. This caused:
1. Throughput calculation to fail (requires counting both start and end tags)
2. Timing marker matching in `parse_test_output.py` to fail (expects start tags to exist)

## Solution
Added start tag output before timing measurement in capturePerf, matching the behavior of the `capture` function:

```javascript
// Print start tag (needed for throughput calculation and timing marker matching)
console.log(`!$######${testStdoutTag}######$!`);
```

## Test plan
- [x] Added test file `runtime/__tests__/capturePerf-start-tags.test.js`
- [x] Test verifies both start and end tags are output
- [x] Test verifies tag format is correct

### DIFF
--- a/packages/codeflash/runtime/__tests__/capturePerf-start-tags.test.js
+++ b/packages/codeflash/runtime/__tests__/capturePerf-start-tags.test.js
@@ -1,0 +1,95 @@
+/**
+ * Test: capturePerf start tag output
+ *
+ * This test verifies that capturePerf outputs START tags before timing,
+ * which is required for:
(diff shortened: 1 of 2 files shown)


Saga4:

PR #1313, opened 2026-02-03
fix: use process.stdout.write for timing markers to bypass Vitest interception

Saga4:
## Summary
- Changes capturePerf timing markers from `console.log` to `process.stdout.write`
- Ensures timing markers are always captured by the parent process

## Problem
Vitest intercepts `console.log` output and may not pass it to the subprocess stdout. This caused timing markers in `capturePerf` to not be captured by the parent process, resulting in "overall summed benchmark runtime is 0" errors.

## Solution
Changed from:
```javascript
console.log(`!######${testStdoutTag}:${durationNs}######!`);
```

To:
```javascript
process.stdout.write(`!######${testStdoutTag}:${durationNs}######!\n`);
```

`process.stdout.write` writes directly to the stdout stream, bypassing any console interceptors.

## Test plan
- [x] Added test file `runtime/__tests__/stdout-write.test.js`
- [x] Test verifies timing markers use `process.stdout.write`
- [x] Test verifies markers are NOT using `console.log`

### DIFF
--- a/packages/codeflash/runtime/__tests__/stdout-write.test.js
+++ b/packages/codeflash/runtime/__tests__/stdout-write.test.js
@@ -1,0 +1,91 @@
+/**
+ * Test: process.stdout.write for timing markers
+ *
+ * This test verifies that capturePerf uses process.stdout.write instead of
+ * console.log for timing markers. This is required because Vitest intercepts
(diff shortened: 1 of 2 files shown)


Saga4:

PR #1314, opened 2026-02-03
fix: add function name fallback for Jest timing marker matching

Saga4:
## Summary
- Adds function name fallback when matching timing markers
- Fixes marker matching when test names are "unknown" in Vitest

## Problem
When Vitest doesn't provide test context (or when tests are configured in certain ways), timing markers may have "unknown" as the test name. The marker matching logic was only matching by test name (group 2), causing all markers to fail matching.

Example failing marker:
```
!$######unknown:unknown:testFunction:1:42_0######$!
```

## Solution
Added fallback to also match by function name (group 3 in the marker), which is always set correctly:

```python
# Before (only test name matching):
matching_starts = [m for m in start_matches if sanitized_test_name in m.group(2)]

# After (test name OR function name matching):
matching_starts = [
    m for m in start_matches
    if sanitized_test_name in m.group(2) or sanitized_func_name == m.group(3)
]
```

## Test plan
- [x] Added test file `tests/test_verification/test_jest_marker_matching.py`
- [x] Tests verify fallback matching works correctly
- [x] Tests verify normal matching still works

### DIFF
--- a/codeflash/cli_cmds/cli.py
+++ b/codeflash/cli_cmds/cli.py
@@ -359,11 +359,13 @@ def _handle_show_config() -> None:
     detected = detect_project(project_root)
 
     # Check if config exists or is auto-detected
-    config_exists, _ = has_existing_config(project_root)
+    config_exists, config_file = has_existing_config(project_root)
     status = "Saved config" if config_exists else "Auto-detected (not saved)"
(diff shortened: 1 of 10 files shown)


Saga4:

PR #1348, opened 2026-02-04
[feat] Java verbose logging

HeshamHM28:


codeflash/optimization/function_optimizer.py
```diff
@@ -146,6 +146,82 @@
     from codeflash.verification.verification_utils import TestConfig
 
 
+def is_verbose_mode() -> bool:
+    """Check if verbose mode is enabled."""
+    return logger.getEffectiveLevel() <= logging.DEBUG
```
KRRT7: we already have this check elsewhere.

codeflash/optimization/function_optimizer.py
```diff
@@ -146,6 +146,82 @@
...
+
+
+def log_code_after_replacement(file_path: Path, candidate_index: int) -> None:
```
KRRT7, line 148: look at how we're logging things for python, we want to do the same, if we don't save to our db then we can't see what went wrong.

KRRT7:
@claude run prek and push

KRRT7:
@claude run prek and push

PR #1376, opened 2026-02-04
fix: support monorepo hoisted dependencies in JS requirements check

mohammedahmed18:
## Problem

The `verify_requirements()` method in JavaScript support only checked for test frameworks (jest/vitest) in the local package's `node_modules`. In monorepos with workspace dependency hoisting (yarn workspaces, pnpm workspaces, etc.), dependencies are often installed at the workspace root instead of in each package.

This caused false positive warnings like:
```
WARNING  JavaScript requirements check found issues:
WARNING    - jest is not installed. Please run 'npm install --save-dev jest' to install it.
```

Even though jest was properly installed at the workspace root.

## Solution

- Check both local `node_modules` and workspace root `node_modules`
- Use existing `_find_monorepo_root()` helper to locate workspace root
- Add debug logging to show where framework was found
- Update docstring to document monorepo support

## Testing

Tested with Budibase monorepo where jest is installed at:
- ❌ `/home/ubuntu/budibase/packages/shared-core/node_modules/jest` (doesn't exist)
- ✅ `/home/ubuntu/budibase/node_modules/jest` (workspace root - now detected)

## Related

This complements the monorepo support already added to:
- jest-runner resolution in loop-runner.js
- Jest benchmarking in test_runner.py

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -54,20 +54,28 @@ jobs:
             PR NUMBER: ${{ github.event.pull_request.number }}
             EVENT: ${{ github.event.action }}
 
-            ## STEP 1: Run pre-commit checks and fix issues
+            ## STEP 1: Run prek and mypy checks, fix issues
 
-            First, run `uv run prek run --from-ref origin/main` to check for linting/formatting issues on files changed in this PR.
+            First, run these checks on files changed in this PR:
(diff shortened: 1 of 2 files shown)


codeflash/languages/javascript/support.py
Saga4, line 1843: ✅ Fixed in latest commit - moved to init_javascript.py as find_node_modules_with_package()

PR #1345, opened 2026-02-04
fix: prevent Maven running all tests + fix TestFile type annotation

mashraf-222:
### Problems Fixed

This PR addresses two related issues that were causing Java E2E optimizations to fail:

**Issue 1: Maven runs ALL tests instead of specific tests**
- When test filter was empty, Maven would run the entire test suite
- Caused unnecessary test execution and parsing errors
- No validation existed to catch this silent failure

**Issue 2: Incorrect type annotation in TestFile model**
- `benchmarking_file_path: Path = None` should be `Optional[Path] = None`
- Caused Pydantic validation errors when path was None
- Prevented proper handling of optional paths

### Root Causes

**Issue 1: Empty test filter**
- `_build_test_filter()` returns empty string when all tests have None paths or fail conversion
- Empty filter → no `-Dtest=` parameter added to Maven command
- Maven default behavior: run ALL tests when no filter specified
- No validation existed to catch this before running Maven

**Issue 2: Invalid type annotation**
- Type annotation `Path = None` is invalid in Pydantic
- Should be `Optional[Path] = None` to allow None values
- Was causing validation errors in test code

### Solutions Implemented

**Issue 1 Fix:**
1. Added validation in `_run_maven_tests()` to raise `ValueError` if filter is empty
2. Added detailed error logging in `_build_test_filter()` to track skipped tests
3. Added warnings when TestFile objects have None paths
4. Prevents silent failure where Maven runs wrong tests

**Issue 2 Fix:**
- Changed `benchmarking_file_path: Path = None` to `Optional[Path] = None`
- Now properly allows None values while maintaining type safety

### Code Changes

**codeflash/languages/java/test_runner.py** (Issue 1):
- Added empty filter validation (raises ValueError)
- Added comprehensive debug logging
- Tracks reasons why tests are skipped
- Provides clear error messages

**codeflash/models/models.py** (Issue 2):
- Fixed type annotation for `benchmarking_file_path`

**codeflash/discovery/discover_unit_tests.py**:
- Included fix to set `tests_project_rootdir` for Java

**tests/test_java_test_filter_validation.py**:
- 4 comprehensive test cases covering both issues
- Tests None path handling
- Tests empty filter detection
- Tests normal operation

### Testing

```bash
pytest tests/test_java_test_filter_validation.py -v
# ✓ 4 passed in 0.09s

# Test cases:
# - test_build_test_filter_with_none_benchmarking_paths ✓
# - test_build_test_filter_with_valid_paths ✓
[...]

PR #1394, opened 2026-02-05
fix: detect and log Java compilation failures explicitly

mashraf-222:
## Problem Fixed

When Maven test execution fails, the error could be due to either:
1. **Compilation errors** - invalid Java syntax (e.g., using reserved keywords)
2. **Test failures** - runtime issues in valid code

Currently, both scenarios produce the same generic error message, making it difficult to diagnose the root cause.

## Root Cause

`_run_maven_tests` combines compilation and test execution in a single Maven command. When Maven returns a non-zero exit code, the code doesn't distinguish between compilation vs runtime failures.

This is particularly problematic when AI-generated tests contain syntax errors (like using "provides" as a class name, which is a Java reserved keyword). The tests fail to compile, but the error appears as a generic "tests failed to run" message.

## Solution Implemented

Added explicit detection of Maven compilation errors by checking output for compilation-specific indicators:
- `[ERROR] COMPILATION ERROR`
- `[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin`
- `compilation failure`
- `cannot find symbol`
- `package .* does not exist`

When compilation errors are detected:
1. Logs ERROR-level message: "Maven compilation failed for {mode} tests"
2. Suggests checking that generated test code is syntactically valid
3. Includes first 50 lines of Maven output for diagnosis

## Code Changes

**File:** `codeflash/languages/java/test_runner.py`
- Lines 1046-1073: Added compilation error detection logic after `subprocess.run`
- Checks both stdout and stderr for compilation error indicators
- Logs detailed error context when compilation fails

## Testing

This change improves error reporting for:
- AI-generated tests with invalid Java syntax
- Tests using reserved keywords as identifiers
- Tests with missing imports or undefined symbols
- Any Maven compilation failure scenario

**Example improved output:**
```
ERROR Maven compilation failed for behavior tests. Check that generated test code is syntactically valid Java. Return code: 1
ERROR Maven compilation error output:
[ERROR] COMPILATION ERROR
[ERROR] /path/to/Test.java:[143,50] cannot find symbol
  symbol:   class provides
  location: package com.example
```

## Impact

This is a detection and logging enhancement with no behavioral changes to test execution. Benefits:
- **Faster diagnosis** - immediately identify compilation vs runtime issues
- **Better debugging** - see actual compilation errors in logs
- **Clearer errors** - understand why tests failed to run

Related to E2E testing where AI-generated tests used Java reserved keywords, causing silent compilation failures.

### DIFF
--- a/.github/workflows/java-e2e-tests.yml
+++ b/.github/workflows/java-e2e-tests.yml
@@ -51,6 +51,12 @@ jobs:
           java -version
           mvn --version
 
+      - name: Build codeflash-runtime JAR
+        run: |
+          cd codeflash-java-runtime
+          mvn clean package -q -DskipTests
+          mvn install -q -DskipTests
+
       - name: Build Java sample project
         run: |
           cd code_to_optimize/java
--- a/.gitignore
+++ b/.gitignore
@@ -168,6 +168,8 @@ cython_debug/
 !tests/test_languages/fixtures/**/pom.xml
 # Allow pom.xml in Java sample project
 !code_to_optimize/java/pom.xml
+# Allow pom.xml in codeflash-java-runtime
+!codeflash-java-runtime/pom.xml
 *.pem
 
 # Ruff cache
--- a/codeflash-java-runtime/pom.xml
+++ b/codeflash-java-runtime/pom.xml
@@ -1,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                             http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.codeflash</groupId>
+    <artifactId>codeflash-runtime</artifactId>
+    <version>1.0.0</version>
+    <packaging>jar</packaging>
+
+    <name>CodeFlash Java Runtime</name>
+    <description>Runtime library for CodeFlash Java instrumentation and comparison</description>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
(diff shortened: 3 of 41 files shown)


mashraf-222:
## Regression Risk: Empty Test Filter Circuit Breaker Removed

This PR adds useful compilation error detection, but it also removes the empty test filter circuit breaker that was introduced in PR #1345 and validated during E2E testing.

### What was removed

On `omni-java`, `_run_maven_tests()` in `codeflash/languages/java/test_runner.py` (lines 1050-1061) has:

```python
    else:
        # CRITICAL: Empty test filter means Maven will run ALL tests
        # This is almost always a bug - tests should be filtered to relevant ones
        error_msg = (
            f"Test filter is EMPTY for mode={mode}! "
            f"Maven will run ALL tests instead of the specified tests. "
            f"This indicates a problem with test file instrumentation or path resolution."
        )
        logger.error(error_msg)
        # Raise exception to prevent running all tests unintentionally
        # This helps catch bugs early rather than silently running wrong tests
        raise ValueError(error_msg)
```

On this PR branch, the entire `else` clause is gone. The `if test_filter:` block at line 1039 no longer has the safety fallback. Additionally, the `logger.debug(f"Added -Dtest={validated_filter} to Maven command")` line was also removed.

### Why this matters

Without the circuit breaker, if `test_filter` is empty (due to a bug in test file instrumentation or path resolution), Maven will silently run **all** tests in the project instead of the targeted subset. This was the exact bug that PR #1345 fixed -- it was discovered during E2E testing where an empty filter caused Maven to execute the entire test suite, producing misleading results and wasting significant time.

### Recommendation

The compilation error detection added by this PR is valuable and can coexist with the empty filter protection. Consider preserving the `else: raise ValueError(...)` block alongside the new compilation error checking logic:

1. Keep the `else` clause on `if test_filter:` that raises `ValueError` on empty filter
2. Keep the `logger.debug` line for validated filter logging
3. The compilation error detection after `subprocess.run` works independently and does not conflict

This way, the PR gains compilation error visibility without losing the safety mechanism that prevents Maven from running all tests when the filter is unexpectedly empty.

PR #1418, opened 2026-02-06
fix: instrument PyTorch nn.Module forward method calls via instance

aseembits93:
## Summary

- Fix instrumentation of PyTorch `nn.Module` forward method when called via instance (e.g., `model(input_data)`)
- Add special handling for the pattern: `model = ClassName(...); model(input_data)` where `model(input_data)` internally calls `forward()`
- Resolves "Ignoring test case that passed but had no runtime" error when optimizing `forward` methods

## Problem

When running `codeflash --function AlexNet.forward`, tests with this pattern weren't being instrumented:

```python
model = AlexNet(num_classes=10)
result = model(input_data)  # calls __call__ which invokes forward()
```

The instrumentation was looking for direct calls to `forward` or `AlexNet`, but `model(input_data)` matched neither.

## Solution

1. Added `collect_instance_variables()` to track variables assigned from class instantiations
2. Modified `find_and_update_line_node()` to wrap calls to instance variables when optimizing `forward` methods
3. Added test case specifically for this PyTorch pattern

## Test plan

- [x] Added `test_pytorch_forward_method_instrumentation` test case
- [x] All existing instrumentation tests pass (19/19)
- [x] Verified with actual codeflash command - runtime is now properly measured

### DIFF
--- a/codeflash/verification/parse_test_output.py
+++ b/codeflash/verification/parse_test_output.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import contextlib
 import os
 import re
 import sqlite3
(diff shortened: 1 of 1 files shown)


KRRT7:
@claude fix the mypy type issues and push

KRRT7:
@claude fix the mypy type issues and push

PR #1422, opened 2026-02-06
fix: resolve test type not found warning for Java tests

aseembits93:
## Summary
- Add fallback logic in `parse_test_output.py` to try `original_file_path` lookup when `instrumented_file_path` lookup fails
- Default to `GENERATED_REGRESSION` test type for Java tests when type cannot be determined
- Add JaCoCo plugin configuration to Java test project pom.xml

## Test plan
- [ ] Run `codeflash --file src/main/java/com/example/Algorithms.java --function fibonacci --verbose` and verify warning is resolved
- [ ] Verify existing Java tests still work correctly

### DIFF
--- a/code_to_optimize/java/pom.xml
+++ b/code_to_optimize/java/pom.xml
@@ -62,6 +62,33 @@
                     </includes>
                 </configuration>
             </plugin>
-        </plugins>
+        
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


code_to_optimize/java/pom.xml
```diff
@@ -62,6 +62,33 @@
                     </includes>
                 </configuration>
             </plugin>
-        </plugins>
+        
+      <plugin>
```
aseembits93, line 66: is this change needed @HeshamHM28 ? this was generated by claude

PR #1496, opened 2026-02-16
[Fix] Java falling tests

HeshamHM28:


codeflash/languages/java/comparator.py
```diff
@@ -90,7 +90,7 @@ def _find_java_executable() -> str | None:
     if platform.system() == "Darwin":
         # Try to extract Java home from Maven (which always finds it)
         try:
-            result = subprocess.run(["mvn", "--version"], capture_output=True, text=True, timeout=10)
+            result = subprocess.run(["mvn", "--version"], capture_output=True, text=True, timeout=10, check=False)
```
misrasaurabh1, line 93: why is this added?

codeflash/languages/java/test_runner.py
```diff
@@ -1428,10 +1430,20 @@ def _path_to_class_name(path: Path) -> str | None:
     if path.suffix != ".java":
         return None
 
-    # Try to extract package from path
-    # e.g., src/test/java/com/example/CalculatorTest.java -> com.example.CalculatorTest
+    path_str = path.as_posix()
     parts = list(path.parts)
 
+    # Try custom source directories first
+    if source_dirs:
```
misrasaurabh1, line 1437: what is this doing? why?

misrasaurabh1 (approved):

codeflash/languages/java/comparator.py
HeshamHM28: The default value for `check` is `False`, so it is not a problem.

codeflash/languages/java/test_runner.py
HeshamHM28: This block handles non-standard Java project layouts. It removes custom source directory prefixes (for example, app/java) from file paths to determine the fully qualified class name. Standard Maven or Gradle layouts (such as src/main/java/ and src/test/java/) are handled by the fallback logic below

PR #1497, opened 2026-02-16
fix: resolve jest-runner from project's node_modules for Jest 30 compatibility

mohammedahmed18:
## Summary
- Fixes Jest 30 loop-runner error: `TypeError: runtime.enterTestCode is not a function`
- Root cause: jest-runner was being loaded from codeflash's node_modules (v29) instead of project's (v30)
- Adds generic recursive search that works with npm, yarn, and pnpm (including non-hoisted deps)

## Changes
- **loop-runner.js**: Replace pnpm-specific resolution with recursive search for `jest-runner/package.json` anywhere in node_modules tree
- **capture.js**: Disable internal looping when external loop-runner is used (Jest handles batching)
- **test_runner.py**: Remove debug print statement
- **parse.py**: Remove unnecessary warning about loop indices

## Test plan
- [x] Tested with novu project (pnpm monorepo, Jest 30.0.5)
- [x] Verified 5 batches run correctly with proper loop indices
- [x] Exit code 0

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


mohammedahmed18:
> Can add unit tests or fix existing on path resolution and loop index

will add unit tests for the js pacakge

PR #1525, opened 2026-02-18
feat: enable coverage check for js/ts

mohammedahmed18:


aseembits93:
@claude why are some ci workflows failing?

aseembits93:
can probably delete the function if it is only returning true @mohammedahmed18 ?

aseembits93:
i see you are doing some kind of inheritance

mohammedahmed18:
> i see you are doing some kind of inheritance

yeah, kevin made a refactor so that each language has its own function optimizer inheriting the base class

PR #1537, opened 2026-02-18
feat: make crosshair-tool optional for Python 3.15+

KRRT7:
## Summary
- Makes `crosshair-tool` dependency conditional on `python_version < '3.15'` in `pyproject.toml`
- Adds a module-level `CROSSHAIR_AVAILABLE` guard in `concolic_testing.py` that skips concolic test generation when the package is absent


KRRT7:
currently crosshair-tool gets built from source on 3.15+

PR #1538, opened 2026-02-18
[Chore] correct speedup formula and formatting in Java runtime comments 

HeshamHM28:


misrasaurabh1:
i dont think the runtime comments works right now, this wont matter until it does

PR #1539, opened 2026-02-19
feat: replace rglob with os.walk and per-language directory pruning

KRRT7:
## Summary
- Replace `Path.rglob` in `get_files_for_language()` with a single `os.walk` pass that prunes excluded directories in-place, avoiding traversal of `.venv`, `node_modules`, `.git`, etc.
- Add `dir_excludes` property to `LanguageSupport` protocol so each language declares its own exclusion patterns (exact names, `prefix*`, `*suffix`)
- Python excludes ~22 patterns (caches, build dirs, VCS); JS/TS excludes 9 (bundler output, framework caches)

## Test plan
- [x] `pytest tests/test_function_discovery.py` — 14 passed
- [x] `pytest tests/test_languages/test_function_discovery_integration.py` — 14 passed
- [x] `ruff check` clean on all changed files

### DIFF
--- a/.codex/config.toml
+++ b/.codex/config.toml
@@ -1,0 +1,4 @@
+[mcp_servers.tessl]
+type = "stdio"
+command = "tessl"
+args = [ "mcp", "start" ]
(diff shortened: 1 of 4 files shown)


KRRT7:
codeflash should optimize this

PR #1542, opened 2026-02-19
ci: add path-based filtering to skip irrelevant workflows on PRs

KRRT7:
## Summary

- JS E2E workflows only trigger when JS-relevant files change (`codeflash/languages/javascript/`, `packages/`, `code_to_optimize/js/`, JS test files)
- Python E2E workflows skip when only JS/npm files change (via `!` negation patterns)
- Python unit test and linter workflows (`unit-tests`, `windows-unit-tests`, `mypy`, `prek`) get path filters scoped to Python-relevant files
- Unit test workflows intentionally keep `code_to_optimize/js/**` included since `test_multi_file_code_replacer.py` depends on JS fixtures
- `push: branches: [main]` and `workflow_dispatch` remain unfiltered

## Test plan

- [ ] Open a PR touching only Python files → JS E2E workflows should not trigger
- [ ] Open a PR touching only `packages/` → Python E2E/unit tests should not trigger
- [ ] Pushes to `main` still trigger everything (no path filter on push)
- [ ] `workflow_dispatch` still works for all workflows

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -12,6 +12,7 @@ on:
       - '!code_to_optimize/js/**'
       - 'pyproject.toml'
       - 'uv.lock'
+      - '.github/workflows/mypy.yml'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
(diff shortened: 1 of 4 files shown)


KRRT7:
actually I'll think of a better way to do this

PR #1543, opened 2026-02-19
feat: implement line-level profiling agent with ASM instrumentation 

HeshamHM28:


mashraf-222 (approved):
## Review: Line-Level Profiling Agent with ASM Instrumentation

### Testing Performed

**All existing tests pass (76/76):**
- 21 unit tests (config generation, class name resolution, JAR location, warmup, boundary conditions, result parsing)
- 11 integration tests (config via JavaSupport, multi-method, nested packages, line contents)
- 44 instrumentation tests (behavior/performance modes, timing markers, edge cases)

**Deep spin-test verification (6 additional tests, all passing):**

| Test | What it validates | Result |
|------|------------------|--------|
| Per-line timing proportionality | 100ms+50ms spin → ~150ms total | 150.1ms total, while-loop 100% of time |
| Raw JSON format | All fields present, correct types, key consistency | All correct, loop hits=1000 exact |
| Multi-method profiling | Two methods (80ms + 40ms) profiled independently | 80.1ms + 40.0ms, ratio 2.00x |
| Self-time attribution | inner(60ms) called from outer(30ms) — outer excludes callee | inner=60.1ms, outer=30.1ms |
| parse_results grouping | Method grouping, sorted lines, str_out formatting | Correctly grouped, tabulate output well-formatted |
| Three-line timing ratios | 30+60+90ms sequential spins, hit counts | 180.1ms total, hits=3 exact |

### Key Observations

**Timing accuracy is excellent** — measured values consistently within 1% of expected spin durations across all tests.

**Self-time attribution works correctly** — the `enterMethod`/`exitMethod` call stack properly excludes callee time from caller lines. When `outer(30ms)` calls `inner(60ms)`, outer's self-time reports ~30ms (not ~90ms).

**Hit counts are exact** — a loop running 1000 iterations reports exactly 1000 hits. A method called 3 times reports exactly 3 hits on its entry line.

**Zero-allocation hot path** — ThreadLocal primitive arrays with no boxing keeps profiling overhead minimal.

**Output format compatibility** — JSON output matches the expected format, `parse_results()` correctly groups by method using config ranges, and `format_line_profile_results()` produces clean tabulate pipe-formatted output.

**Note:** The `.m2` cached JAR needs rebuilding (`mvn clean package -DskipTests` + `mvn install`) since this PR adds `Premain-Class` to the manifest. The bundled JAR in `resources/` is included in the diff so CI should be fine.

LGTM — the profiler produces accurate, well-structured output and the Python integration layer correctly parses and formats it.

PR #1545, opened 2026-02-19
version bump

misrasaurabh1:


KRRT7:
@misrasaurabh1 was wanting to set up trusted published for future automatic releases: #1536

PR #1552, opened 2026-02-19
fix: Java E2E optimization pipeline issues - 64% failure reduction, 10-20x speedup

mashraf-222:
# Fix Java E2E Optimization Pipeline Issues

## Summary

This PR addresses critical bugs in the Java E2E optimization pipeline discovered during a comprehensive bug hunting session with the aerospike-client-java project. The fixes resolve 64% of test failures and achieve 10-20x performance improvements in test execution.

**Impact:** These changes fix 4 major bugs affecting Java optimizations, with 2 fully resolved and 2 partially resolved. One pre-existing bug (Bug 10) was discovered during verification and requires a separate fix.

### 🔴 Bug #7: JUnit Version Detection Failure (64% of all failures) - FULLY FIXED

**Problem:**
- The pom.xml parser only checked the `<dependencies>` section, missing dependencies declared in `<dependencyManagement>`
- This caused JUnit 4 projects to be incorrectly identified as JUnit 5
- Generated incompatible test code leading to compilation failures

**Solution:**
- Modified `_detect_test_deps_from_pom()` in `codeflash/languages/java/config.py` to parse both `<dependencies>` and `<dependencyManagement>` sections
- Changed default fallback from JUnit 5 to JUnit 4 (more common in legacy projects)

**Code Changes:**
```python
# codeflash/languages/java/config.py
def check_dependencies(deps_element, ns):
    """Check dependencies element for test frameworks."""
    # Now checks both <dependencies> and <dependencyManagement>
    for dependency in deps_element.findall(".//maven:dependency", ns):
        # ... detection logic ...
```

**Result:** 100% accurate JUnit version detection, eliminating 28 out of 42 test failures

---

### 🔴 Bug #3: Direct JVM Execution Failure - FULLY FIXED

**Problem:**
- Tests couldn't run directly with JVM, always fell back to Maven
- Maven execution took 5-10 seconds per test loop
- Multi-module project classpaths weren't properly constructed

**Solution:**
- Implemented JUnit 4 vs JUnit 5 detection at runtime
- Use `org.junit.runner.JUnitCore` for JUnit 4, `ConsoleLauncher` for JUnit 5
- Fixed classpath construction for multi-module Maven projects

**Code Changes:**
```python
# codeflash/languages/java/test_runner.py
# Detect JUnit version
is_junit4 = check_for_junit4_in_classpath()

if is_junit4:
    cmd = [java, "-cp", classpath, "org.junit.runner.JUnitCore", *test_classes]
else:
[...]

PR #1580, opened 2026-02-20
fix: Java E2E pipeline — direct JVM benchmarking, JUnit detection, and instrumentation fixes

mashraf-222:
## Summary

This PR fixes 5 out of 9 critical bugs found during Java E2E optimization testing with the aerospike-client-java project. It builds on and supersedes the previous PR #1552, which was reverted due to accumulated complexity. This branch was rebuilt incrementally with E2E validation at each step.

**Key results:**
- Direct JVM benchmarking now works end-to-end (was always falling back to Maven)
- JUnit version detection fixed for multi-module Maven projects (64% of original failures)
- Test execution overhead reduced from ~5-10s to ~0.65s per benchmark loop
- Framework detection calls reduced from 303,000+ to 6 per optimization run
- Test method names correctly stored in SQLite for behavior tests

### Bug #7 (Critical): JUnit Version Detection Failure — 64% of all failures

**Problem:** The pom.xml parser only checked `<dependencies>`, missing test dependencies declared in `<dependencyManagement>` (common in multi-module Maven projects like Aerospike). This caused JUnit 4 projects to be incorrectly identified as JUnit 5, generating incompatible test code.

**Root cause:** `_detect_test_deps_from_pom()` in `config.py` didn't parse `<dependencyManagement>` sections or check submodule pom.xml files.

**Fix:** 
- Refactored `_detect_test_deps_from_pom()` to parse both `<dependencies>` and `<dependencyManagement>` sections using a shared `check_dependencies()` helper
- Added recursive checking of submodule pom.xml files (`test/`, `tests/`, `src/test/`, `testing/`)
- Changed default fallback from JUnit 5 to JUnit 4 (more common in legacy projects)
- Added debug logging for framework detection decisions

### Bug #3 (High): Direct JVM Execution Always Failing

**Problem:** Direct JVM benchmarking never worked — it always fell back to Maven. Three separate issues:

1. **ConsoleLauncher not on classpath:** `junit-platform-console-standalone` is a separate artifact not included in the normal dependency tree. `mvn dependency:build-classpath` doesn't output it, so `ConsoleLauncher` class was not found at runtime.

2. **False JUnit 4 detection:** The old code ran `java -cp ... JUnitCore -version` to detect JUnit 4. But JUnit 5 projects include JUnit 4 classes via `junit-vintage-engine`, so this check always returned true. JUnitCore then ran against JUnit 5 tests, found 0 tests, and triggered the Maven fallback.

3. **Missing perf_stdout:** After direct JVM execution, `perf_stdout` wasn't populated from the subprocess result, so the throughput calculation pipeline had no timing data.

**Fixes:**
- Added `_find_junit_console_standalone()` to locate the JAR in `~/.m2/repository`. If not present, downloads via `mvn dependency:get`. Appends to classpath in `_get_test_classpath()`.
- Replaced subprocess-based JUnit detection with classpath string inspection: checks for `junit-jupiter`, `junit-platform`, or `console-standalone` in the classpath. This is deterministic, instant, and not fooled by vintage-engine compatibility classes. Addresses [PR #1552 review comment](https://github.com/codeflash-ai/codeflash/pull/1552#discussion_r2829540508) about avoiding per-execution detection overhead.
- Added multi-module classpath support: includes `target/classes` from sibling modules for projects where test modules depend on other modules.
- Fixed `perf_stdout` capture in `parse_test_output.py` to extract stdout from subprocess results for Java performance tests.

### Bug #6 (High): Instrumentation Breaking Complex Expressions

**Problem:** Timing markers were inserted inside cast expressions, ternary operators, array access, and other complex expressions, causing "not a statement" compilation errors (e.g., `(Long)list.get(2)` broken by instrumentation).

**Fix:**
- Added `_is_inside_complex_expression()` that walks the tree-sitter AST upward to detect problematic parent types: `cast_expression`, `ternary_expression`, `array_access`, `binary_expression`, `unary_expression`, `parenthesized_expression`, `instanceof_expression`.
- Stops at statement boundaries to avoid false positives.
- Both behavior and timing instrumentation now skip calls inside complex expressions.
- Added `in_complex` flag to the call collection pipeline.

### Bug #2 (Medium-High): Extremely Slow rglob Calls

**Problem:** `resolve_test_file_from_class_path()` was called for every timing marker without caching, causing 43+ `rglob` calls per optimization on large projects.

**Fix:**
- Added `_test_file_path_cache: dict[tuple[str, Path], Path | None]` module-level cache
- Caches both positive and negative lookup results
- All resolution paths (direct match, rglob fallback, Java-specific) store results in cache

### Pre-existing Bug: TestConfig.test_framework Uncached Property

**Problem:** `TestConfig.test_framework` was an uncached `@property` that re-detected the framework on every access. During test result parsing, each test result accessed this property, causing 303,000+ calls to `_detect_test_deps_from_pom()` and 2.3M lines of output per optimization run. This bug exists on `omni-java` and was not introduced by this branch.

**Fix:**
- Added `_test_framework: Optional[str] = None` cache field to `TestConfig`
- Property now returns cached value after first detection
- Also changed default fallback in `_detect_java_test_framework()` from `junit5` to `junit4`

### Behavior Test Method Name Fix

**Problem:** SQLite `setString(3, ...)` in behavior instrumentation used a hardcoded `"{class_name}Test"` string instead of the actual test method name. This meant all test results mapped to the same identifier, losing per-test granularity.

**Fix:**
- Added `_extract_test_method_name()` with two regex patterns: `_METHOD_SIG_PATTERN` (full Java method signature) and `_FALLBACK_METHOD_PATTERN` (simple name extraction)
- Each instrumented test method now declares `String _cf_test{N} = "{methodName}"` and uses it in the SQLite insert
- Added `bisect` optimization for `_byte_to_line_index()` (O(log n) vs O(n) per call)

## Code Changes

| File | Lines | Description |
|------|-------|-------------|
| `codeflash/languages/java/config.py` | +71/-27 | Parse `<dependencyManagement>`, check submodule pom.xml files, change default to JUnit 4 |
| `codeflash/languages/java/test_runner.py` | +139/-37 | `_find_junit_console_standalone()`, classpath string JUnit detection, multi-module classpath, `--add-opens` for JUnit 4 path |
| `codeflash/languages/java/instrumentation.py` | +76/-11 | `_is_inside_complex_expression()`, `_extract_test_method_name()`, `in_complex` flag, `bisect` for line index, correct SQLite test name |
| `codeflash/verification/verification_utils.py` | +12/-4 | Cache `test_framework` property, change default fallback to JUnit 4 |
| `codeflash/verification/parse_test_output.py` | +28/-2 | Path resolution cache, `perf_stdout` capture for Java performance tests |
| `tests/test_languages/test_java/test_instrumentation.py` | +16/-8 | Updated 8 expected strings for `_cf_test` variable and `setString(3, ...)` |

### Other files with minor formatting changes (from pre-commit):
- `codeflash/cli_cmds/console.py`, `codeflash/cli_cmds/logging_config.py`, `codeflash/context/code_context_extractor.py`, `codeflash/languages/java/context.py`, `codeflash/optimization/function_optimizer.py`, `codeflash/verification/parse_line_profile_test_output.py`

### E2E Validation (Fibonacci — JUnit 5, single-module)
```bash
cd code_to_optimize/java/
CODEFLASH_CFAPI_SERVER=local CODEFLASH_AIS_SERVER=local \
  uv run codeflash --file src/main/java/com/example/Fibonacci.java --function fibonacci --verbose --no-pr
```
- JUnit 5 correctly detected on every invocation
- **Zero Maven fallbacks** — all benchmark loops via direct JVM ConsoleLauncher
- 5,817% speedup found and accepted
- Loop times: ~0.65s (was ~5-10s with Maven fallback)
- Framework detection: 6 calls (was 303,000+)

### E2E Validation (BubbleSort — exercises instrumentation + measurement)
```bash
uv run codeflash --file src/main/java/com/example/BubbleSort.java --function bubbleSort --verbose --no-pr
```
- JUnit 5 correctly detected, zero Maven fallbacks
- 11 optimization candidates tested, all correctness tests passed
- No optimization found (expected — BubbleSort near-optimal for test inputs)
- Pipeline completed cleanly with no errors

### Unit Tests
- All 41 Java instrumentation tests pass (`tests/test_languages/test_java/test_instrumentation.py`)
- Updated 8 expected strings for `_cf_test` variable and `setString(3, _cf_test{N})`

## Performance Impact

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Benchmark loop time | ~5-10s (Maven) | ~0.65s (direct JVM) | 8-15x faster |
| Framework detection calls/run | 303,000+ | 6 | 50,000x fewer |
| Verbose output lines/run | 2,300,000+ | ~13,000 | 177x smaller |
| Maven fallback rate | 100% | 0% | Eliminated |
| rglob calls per function | 43+ uncached | 1 + cache hits | Eliminated redundancy |

## Known Issues Not Addressed

- **Bug #1**: AI generates implicit int-to-byte conversions — requires AI service fix
- **Bug #5**: JaCoCo XML parsed repeatedly — low priority
- **Bug #8**: AI response truncation — requires AI service fix
- **Bug #9**: Generated test file cleanup — requires architecture change
- **Aerospike Surefire `<includes>` bypass**: Behavior mode still uses Maven (not direct JVM). Aerospike's custom Surefire `<includes>` config blocks `-Dtest=...` filtering. Plan documented but not implemented in this PR.

## Relationship to PR #1552

This PR supersedes #1552. The original branch had 12 commits that were reverted due to accumulated debug logging and regressions. This branch was reset to the last known-good commit and rebuilt with:
- No debug/investigation commits
- Each fix independently validated via E2E
- Improved JUnit detection (classpath string check vs subprocess probing)
- ConsoleLauncher classpath fix (not in #1552)
- TestConfig caching fix (not in #1552)

### DIFF
--- a/codeflash/languages/java/instrumentation.py
+++ b/codeflash/languages/java/instrumentation.py
@@ -840,7 +840,9 @@ def _add_timing_instrumentation(source: str, class_name: str, func_name: str) ->
         assignment = f"{name_text} = {value_text};"
         return hoisted, assignment
 
-    def build_instrumented_body(body_text: str, next_wrapper_id: int, base_indent: str, test_method_name: str = "unknown") -> tuple[str, int]:
+    def build_instrumented_body(
+        body_text: str, next_wrapper_id: int, base_indent: str, test_method_name: str = "unknown"
+    ) -> tuple[str, int]:
         body_bytes = body_text.encode("utf8")
         wrapper_bytes = _TS_BODY_PREFIX_BYTES + body_bytes + _TS_BODY_SUFFIX.encode("utf8")
         wrapper_tree = analyzer.parse(wrapper_bytes)


tests/test_languages/test_java/test_instrumentation.py
```diff
@@ -256,6 +257,7 @@ def test_instrument_behavior_mode_assert_throws_expression_lambda(self, tmp_path
         String _cf_outputFile1 = System.getenv("CODEFLASH_OUTPUT_FILE");
         String _cf_testIteration1 = System.getenv("CODEFLASH_TEST_ITERATION");
         if (_cf_testIteration1 == null) _cf_testIteration1 = "0";
+        String _cf_test1 = "testNegativeInput_ThrowsIllegalArgumentException";
```
misrasaurabh1, line 260: i still think this identifier isn't enough. we need a unique identifier that if we see a given marker, we know a completely unique identity for that run. so uniqueness for a test invocation is defined by
test_file -> test_class -> test_function -> test_line_node ->run_counter.
This is largely how python instrumentation works. currently i see that the module name isn't unique (since it lacks the __perfinstrumented section - it may be fine to skip this but then look at what the python instrumentation does for a reference implementation).
There are other things we will fix shortly, the line numbers (needed for runtime annotations on the tests we display on PRs). will test more

misrasaurabh1:
we would also need the same test function name in the performance instrumentation

misrasaurabh1:
in behavior instrumentation we are still not printing the test class +fn name in stdout markers
```
int _cf_loop5 = Integer.parseInt(System.getenv("CODEFLASH_LOOP_INDEX"));
        int _cf_iter5 = 5;
        String _cf_mod5 = "CryptoTest";
        String _cf_cls5 = "CryptoTest";
        String _cf_fn5 = "computeDigest";
        String _cf_outputFile5 = System.getenv("CODEFLASH_OUTPUT_FILE");
        String _cf_testIteration5 = System.getenv("CODEFLASH_TEST_ITERATION");
        if (_cf_testIteration5 == null) _cf_testIteration5 = "0";
        String _cf_test5 = "testComputeDigest_EmptySetNameAndEmptyStringKey_Returns20ByteDigest";
        System.out.println("!$######" + _cf_mod5 + ":" + _cf_cls5 + ":" + _cf_fn5 + ":" + _cf_loop5 + ":" + _cf_iter5 + "######$!");
        
```
`_cf_test5` is not present in the stdout

mashraf-222:
## Update: Fixed all outstanding issues from verification

Four additional commits pushed to address issues found during E2E verification and review feedback:

### 1. `fix: implement Java process_review methods to prevent crash after optimization`
- **Problem:** After a successful optimization, `process_review` crashed with `AttributeError: 'NoneType' has no attribute 'generated_tests'` because the base class stubs for `remove_test_functions_from_generated_tests()` and `add_runtime_comments_to_generated_tests()` return `None`.
- **Fix:** Added full implementations of both methods + `_build_runtime_map()` helper to `JavaSupport`, following the JavaScript pattern.

### 2. `fix: add test method name to Java stdout markers for unique identification`
- **Problem:** Java stdout markers used format `mod:cls:fn:loop:iter` but lacked the test method name, making per-test filtering impossible from stdout alone. (Addresses @misrasaurabh1's review comment about marker uniqueness.)
- **Fix:** Markers now emit `mod:cls.testMethod:fn:loop:iter` for both behavior and timing instrumentation. Parser updated to extract test method name from the combined `class.test` field. All 41 instrumentation tests updated and passing.

### 3. `fix: clear test file path cache between optimization iterations in --all mode`
- **Problem:** Module-level `_test_file_path_cache` persisted across `--all` mode iterations. Negative cache entries (`None`) could mask test files generated in later iterations.
- **Fix:** Added `clear_test_file_path_cache()` called after each optimization cleanup in the finally block.

### 4. `chore: log debug message when JUnitCore ignores reports_dir parameter`
- **Problem:** JUnitCore branch silently ignores `reports_dir` parameter, which could be confusing during debugging.
- **Fix:** Added a debug log when `reports_dir` is provided but JUnitCore is being used.

### E2E Validation Results

All changes validated with Fibonacci + Aerospike E2E runs after each phase:

| Run | Target | Result |
|-----|--------|--------|
| Phase 2 - Fibonacci (`--no-pr`) | `fibonacci` | **PASSED** — 476,142% speedup, no `AttributeError` crash |
| Phase 2 - Aerospike | `Utf8.encodedLength` | **PASSED** — No optimization found (expected), pipeline clean |
| Phase 3+4 - Fibonacci (`--no-pr`) | `fibonacci` | **PASSED** — 153,136% speedup, `mark-as-success` |
| Phase 3+4 - Aerospike | `Utf8.encodedLength` | **PASSED** — 23% speedup found, [PR ](https://github.com/mashraf-222/aerospike-client-java/pull/13)created |

mashraf-222:
Hey @misrasaurabh1 — noticed your `7d329adc` commit that added `remove_test_functions_from_generated_tests()` and `add_runtime_comments_to_generated_tests()` to `JavaSupport`. We were working on the same fix independently!

Your `remove_test_functions_from_generated_tests()` implementation was identical to ours, so that merged cleanly. For `add_runtime_comments_to_generated_tests()`, we built on top of your stub by adding a full implementation that:

- Builds runtime maps from `InvocationId`-keyed data (via a `_build_runtime_map()` helper, following the JS pattern)
- Calls the existing `self.add_runtime_comments()` to inject runtime comments into each generated test's source

So instead of returning tests unchanged, it now actually decorates them with original vs optimized runtime info — same as JS does. Let us know if you have any concerns with that approach.

mashraf-222:
> in behavior instrumentation we are still not printing the test class +fn name in stdout markers [...] `_cf_test5` is not present in the stdout

Addressed in commit `1cdbac4b` (`fix: add test method name to Java stdout markers for unique identification`). Both behavior and timing markers now include the test method name:

```java
// Before:
System.out.println("!$######" + _cf_mod5 + ":" + _cf_cls5 + ":" + _cf_fn5 + ":" + _cf_loop5 + ":" + _cf_iter5 + "######$!");

// After:
System.out.println("!$######" + _cf_mod5 + ":" + _cf_cls5 + "." + _cf_test5 + ":" + _cf_fn5 + ":" + _cf_loop5 + ":" + _cf_iter5 + "######$!");
```

The parser in `parse_test_output.py` was updated to extract the test method name from the `class.testMethod` field. All 41 instrumentation tests pass, and E2E validated with both Fibonacci and Aerospike runs.

PR #1618, opened 2026-02-20
feat: add --agent mode flag

KRRT7:
## Summary
- Adds `--agent` CLI flag for Claude / agents: suppresses Rich UI, emits plain-text/XML output, skips checkpoint and concolic tests, implies `--no-pr` and `--worktree`.


misrasaurabh1:
why dont you use the word --subagent? agent mode within codeflash means something else

PR #1649, opened 2026-02-24
perf: fix two performance bottlenecks

aseembits93:
## Summary

- **Cache normalized imports in optimizer**: `prepare_module_for_optimization` now caches normalized imported modules in `self.normalized_imports_cache`, eliminating redundant file I/O + AST parse/unparse when multiple functions share helper modules
- **Move conditional imports to module level in comparator**: All 11 conditional library imports (jax, numpy, torch, pandas, etc.) moved from inside the recursive `comparator()` function body to module level, wrapped in `try/except ImportError` for robustness — eliminates per-call import machinery overhead

## Test plan

- [x] `uv run pytest tests/test_comparator.py -x -q` — 169 passed, 6 skipped
- [x] Verified all three modules import cleanly
- [x] Verified `comparator()` works on basic types after import changes
- [x] Verified `HAS_*` flags correctly fall back to `False` when import fails (e.g., numba with incompatible numpy)


codeflash/api/cfapi.py
```diff
@@ -77,9 +79,9 @@ def make_cfapi_request(
         if method.upper() == "POST":
             json_payload = json.dumps(payload, indent=None, default=pydantic_encoder)
             cfapi_headers["Content-Type"] = "application/json"
-            response = requests.post(url, data=json_payload, headers=cfapi_headers, timeout=60)
+            response = cfapi_session.post(url, data=json_payload, headers=cfapi_headers, timeout=60)
```
misrasaurabh1: no need

PR #1650, opened 2026-02-24
fix: detect attribute-referenced methods as used in unused helper detection

aseembits93:
## Summary
- `detect_unused_helper_functions` only walked `ast.Call` nodes to find which helpers were "used" by the optimized entrypoint
- Methods referenced via attribute assignment (e.g., `self._parse1 = self._parse_literal` — a common callback/state-machine pattern) were missed, causing them to be falsely classified as unused and reverted to their original unoptimized code
- Added an `elif` branch that also walks `ast.Attribute` nodes with `ast.Name` values, using the same qualified-name and import-mapping logic as the call-based detection

## Test plan
- [x] 3 new tests in `tests/test_mock_candidate_replacement.py` (mock candidate with PSBaseParser class methods used as callbacks)
- [x] 17 existing tests in `tests/test_unused_helper_revert.py` all pass unchanged

### DIFF
--- a/tests/test_mock_candidate_replacement.py
+++ b/tests/test_mock_candidate_replacement.py
@@ -415,6 +415,209 @@ class PSBaseParser:
 ```
 '''
 
+EXPECTED_OUTPUT = '''\
+import contextlib
+from typing import BinaryIO, TypeVar, Union
+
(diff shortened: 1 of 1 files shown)


KRRT7:
@aseembits93 is this good to go?

aseembits93:
@KRRT7 works

PR #1665, opened 2026-02-25
fix: use correct subagent flag in checkpoint guard

aseembits93:
## Summary
- The checkpoint guard in `optimizer.py` referenced `args.agent`, which is never defined on the CLI args namespace — the actual flag is `--subagent` (`args.subagent`).
- Since `getattr(self.args, "agent", False)` always returned `False`, the condition was effectively just `self.args.all`, meaning checkpointing was never skipped in subagent mode as intended.

## Test plan
- [ ] Verify `--all --subagent` skips checkpoint creation
- [ ] Verify `--all` without `--subagent` still creates checkpoint

### DIFF
--- a/codeflash-benchmark/codeflash_benchmark/version.py
+++ b/codeflash-benchmark/codeflash_benchmark/version.py
@@ -1,2 +1,2 @@
 # These version placeholders will be replaced by uv-dynamic-versioning during build.
-__version__ = "0.20.2"
+__version__ = "0.3.0"


aseembits93:
@KRRT7 fixed

KRRT7:
@aseembits93 big jump?

aseembits93:
> @aseembits93 big jump?

i only see a single line of change, i think it was out of sync with main before?

KRRT7:
oops, I misread the update

PR #1666, opened 2026-02-25
feat: surface subagent optimization diffs in IDE's native diff view

aseembits93:
## Summary
- Updated subagent mode's `<action>` instructions to tell Claude Code to use the `Write` tool instead of `AskUserQuestion`
- When Claude Code writes optimized code to files, the connected IDE (VS Code, Cursor, Windsurf) surfaces the changes in its native diff viewer
- Users can accept or reject optimizations through the IDE's diff UI rather than reading a text diff in the chat

## Test plan
- [ ] Run `codeflash --subagent` and verify the XML output contains the updated `<action>` block
- [ ] Invoke via Claude Code in an IDE-connected session and confirm the diff appears in the IDE's diff viewer

### DIFF
--- a/codeflash/cli_cmds/console.py
+++ b/codeflash/cli_cmds/console.py
@@ -469,9 +469,17 @@ def subagent_log_optimization_result(
             xml.append(f'  <optimized-code file="{escape(str(path))}">{escape(new_code[path])}</optimized-code>')
     xml.append("  <action>")
     xml.append("    First, present a short summary: function name, performance improvement, and explanation.")
-    xml.append("    Then, for each optimized-code element, use the Write tool to write the optimized code to the file.")
-    xml.append("    This will surface the diff in the user's IDE (VS Code, Cursor, Windsurf, etc.) for review.")
-    xml.append("    The user can accept or reject the changes through the IDE's native diff view.")
+    xml.append("    If you are running inside an IDE (VS Code, Cursor, Windsurf, etc.):")
+    xml.append("      For each optimized-code element, use the Write tool to write the optimized code to the file.")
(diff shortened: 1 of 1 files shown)


aseembits93:
@KRRT7 the existing style of diff would collapse if the diff is large enough, if we rely on claude's native style of diff, it handles it slightly better imo

KRRT7:
personally, I've been using Claude Code + Warp which has a built in diff view, which allows me to view them without ever using my IDE, so I keep that open less and less, and the creator of Claude Code has said he and a lot of engineers have stopped using IDEs entirely, this will break that flow, so you should look into how to support both

aseembits93:
@KRRT7 in the current setup are you able to amend the diff, make changes which are not part of the diff?

aseembits93:
https://github.com/anthropics/claude-code/issues/6735 it's a feature request unfortunately with CC . once CC presents a diff, the user cannot manually modify it, the user can only instruct claude in english to make the necessary change

aseembits93:
https://github.com/anthropics/claude-code/issues/27708 this one too

KRRT7:
i prompt it to make the changes, or i make the changes thru nano / warp and tell claude i've made the changes and it auto shows me the changes

the workflow with Claude Code has largely changed to just prompting Claude to make the changes itself 

@misrasaurabh1 thoughts?

misrasaurabh1:
I would need to see how this looks like compared the the alternative - at a very high level I like the native claude diff handling since it makes people feel more comfortable with the new code.

KRRT7:
@aseembits93 I guess we can.merge?

aseembits93:
@KRRT7 let's merge, will iterate more with new PRs

misrasaurabh1:
i am merging so that i can try with the latest version

PR #1686, opened 2026-02-27
perf: optimize tracer hot path with string-based path ops and caching

KRRT7:
## Summary
- Cache path resolution and validity checks to avoid redundant `os.path.realpath` and `os.path.exists` calls on repeated function entries
- Replace `Path.is_relative_to` and `Path.resolve` with `os.path.realpath` + `str.startswith` in the hot path for lower overhead
- Defer `Path` construction to first-seen functions only, keeping the per-call path as string-based
- Restore `tests-root` to `tests` in pyproject.toml

## Test plan
- Existing tracer tests should pass
- Profile with a real project to verify reduced overhead in `profile_callback`


KRRT7:
this was causing 10-15% overhead

PR #1689, opened 2026-02-27
refactor: consolidate Python function discovery to CST path only

KRRT7:
## Summary
- Remove the AST-based Python discovery path (`find_functions_with_return_statement`, `_find_all_functions_in_python_file`, `function_has_return_statement`, `function_is_a_property`, and associated constants)
- Route Python through the unified `_find_all_functions_via_language_support` like all other languages
- Fix `FunctionVisitor` to skip nested functions and exclude `@property`/`@cached_property` decorated functions
- Let parse errors propagate from `PythonSupport.discover_functions` so invalid files correctly return `{}`

## Test plan
- [x] `uv run pytest tests/test_function_discovery.py -x -q` — 23 passed
- [x] `uv run pytest tests/test_async_function_discovery.py -x -q` — all passed
- [x] `uv run ruff check` / `ruff format` — clean

### DIFF
--- a/codeflash/discovery/functions_to_optimize.py
+++ b/codeflash/discovery/functions_to_optimize.py
@@ -68,9 +68,9 @@ class ReturnStatementVisitor(cst.CSTVisitor):
 class FunctionVisitor(cst.CSTVisitor):
     METADATA_DEPENDENCIES = (cst.metadata.PositionProvider, cst.metadata.ParentNodeProvider)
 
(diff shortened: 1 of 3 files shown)


KRRT7:
@claude Test results: 2463 passed, 13 failed, 57 skipped. The 13 failures are pre-existing (same failures occur on main). what failures are those?

KRRT7:
@claude is the JUnit XML parsing logic actually being used for JS / TS?

PR #1733, opened 2026-03-03
scaffold java support

KRRT7:


aseembits93:
@KRRT7 merging with main or omni-java?

KRRT7:
@aseembits93 with main and from there incrementally merging, the python / javascript part of the codebase are very ahead and there's lots of tricky areas if it start merging main into omni-java, this is much easier for me

PR #1745, opened 2026-03-03
fix: add pre-flight module import check to exit early before API calls

aseembits93:
## Summary
- Adds a find_spec import validation check in `can_be_optimized()` that runs **before** any expensive API calls (test generation + optimization)
- If the import fails due to a wrong `module-root`, automatically infers the correct value from the `__init__.py` chain, updates `pyproject.toml`, and retries
- If the import fails due to a missing dependency (not a module-root issue), fails early with a clear error message instead of wasting API credits

## Test plan
- [ ] Set `module-root` to an incorrect value in `pyproject.toml` and run `codeflash --file <file> --function <func>` — verify it auto-corrects and updates `pyproject.toml`
- [ ] Remove a dependency import and verify the pre-flight check catches it before API calls
- [ ] Run with a correct `module-root` and verify no regression — optimization proceeds normally

### DIFF
--- a/codeflash/languages/python/function_optimizer.py
+++ b/codeflash/languages/python/function_optimizer.py
@@ -9,7 +9,11 @@ import tomlkit
 
 from codeflash.cli_cmds.cli import project_root_from_module_root
 from codeflash.cli_cmds.console import console, logger
-from codeflash.code_utils.code_utils import infer_module_root_from_file, module_name_from_file_path, validate_module_import
+from codeflash.code_utils.code_utils import (
+    infer_module_root_from_file,
+    module_name_from_file_path,
+    validate_module_import,
+)
 from codeflash.code_utils.config_consts import TOTAL_LOOPING_TIME_EFFECTIVE
 from codeflash.code_utils.config_parser import find_pyproject_toml
 from codeflash.either import Failure, Success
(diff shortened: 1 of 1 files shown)


KRRT7:
a full subprocess call is far more expensive than how we check if something is installed in the comparator:
https://github.com/codeflash-ai/codeflash/blob/main/codeflash/verification/comparator.py#L21

you can use that for this

aseembits93:
this pr is still in progress

aseembits93:
@claude is the windows unittest failure related to this pull request?

aseembits93:
@claude is the windows unittest failure related to this pull request?

aseembits93:
@claude fix the failing `tests/test_languages/test_javascript_test_runner.py::TestBundledJestReporter::test_reporter_produces_valid_junit_xml`

codeflash/optimization/function_optimizer.py
```diff
@@ -555,6 +611,19 @@ def can_be_optimized(self) -> Result[tuple[bool, CodeOptimizationContext, dict[P
                     f"Cannot optimize without tests when --no-gen-tests is set."
                 )
 
+        # Pre-flight: verify module-root consistency and importability before expensive API calls
```
KRRT7: instead of implementing it in the function optimizer here, can you find a good spot for it in https://github.com/codeflash-ai/codeflash/blob/main/codeflash/languages/python/function_optimizer.py?

KRRT7 (changes requested):

codeflash/optimization/function_optimizer.py
KRRT7: if pre-flight doesn't exist in the language base add it

PR #1808, opened 2026-03-10
fix: use runtime Jest config for test files outside project root

mohammedahmed18:
## Summary

- Replaces CLI `--roots` flags with a runtime wrapper Jest config that extends the base config with additional `roots` and `testMatch` entries
- Fixes test discovery in monorepo setups where codeflash-generated test files live outside the project root (`<rootDir>`)
- Applies the runtime config consistently across behavioral, benchmarking, and line profile test runners

## Test plan

- [ ] Verify Jest behavioral tests pass when test files are inside the project root (no runtime config created)
- [ ] Verify Jest tests work in monorepo setups where test files are generated outside the project root
- [ ] Confirm the runtime config file is cleaned up after test execution

### DIFF
--- a/tests/test_languages/test_javascript_test_runner.py
+++ b/tests/test_languages/test_javascript_test_runner.py
@@ -8,28 +8,23 @@ import pytest
 
 
 class TestJestRootsConfiguration:
-    """Tests for Jest --roots flag handling."""
+    """Tests for Jest runtime config creation when test files are outside the project root."""
 
-    def test_behavioral_tests_adds_roots_for_test_directories(self):
-        """Test that run_jest_behavioral_tests adds --roots for test directories."""
-        from codeflash.languages.javascript.test_runner import run_jest_behavioral_tests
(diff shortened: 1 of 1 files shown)


codeflash/languages/javascript/test_runner.py
```diff
@@ -339,6 +339,73 @@ def _create_codeflash_jest_config(
         return None
 
 
+def _create_runtime_jest_config(base_config_path: Path | None, project_root: Path, test_dirs: set[str]) -> Path | None:
```
Saga4, line 342: So how is it working now when generated tests are outside project root?

mohammedahmed18: it's the same the only difference is we are merging now the roots instead of overriding them by --roots. in strapi overriding the root dir (root of monorepo) to become <rootDir>/tests (our tests dir) caused jest to fail because of strapi jest config assumed that rootDir is the monorepo root like here `<rootDir>/packages/**/*.{js,ts}`

PR #1815, opened 2026-03-11
fix: ensure worktree cleanup on SIGTERM, SIGHUP, and atexit

aseembits93:
## Summary

Ensures the git worktree is always cleaned up when running with `--worktree`, regardless of how the process exits:

- **Signal handlers** for `SIGTERM`, `SIGHUP`, `SIGQUIT`, and `SIGPIPE` — covers `kill`, Docker/systemd stop, terminal close, broken pipe, and `Ctrl+\`
- **`atexit` handler** as a safety net for unhandled exceptions and `sys.exit()` from elsewhere
- **`KeyboardInterrupt`** (`Ctrl+C`) — already handled, unchanged
- **Stale worktree cleanup at startup** — writes a `.codeflash.pid` file when creating a worktree; on next run, scans the worktrees directory and removes any whose owning process is dead (handles `SIGKILL` / `kill -9`)

Signal handlers and atexit registration are restored/unregistered in a `finally` block to avoid side effects outside the optimizer run.

## Test plan

- [x] Existing `tests/test_worktree.py` passes
- [x] Ruff lint passes
- [x] Manual: run with `--worktree`, send `SIGTERM` — verify worktree directory is removed
- [x] Manual: run with `--worktree`, close terminal (`SIGHUP`) — verify worktree directory is removed
- [x] Manual: run with `--worktree`, `Ctrl+C` — verify worktree directory is removed (existing behavior)
- [x] Manual: run with `--worktree`, `kill -9` — verify worktree is cleaned up on next run
- [x] Manual: run two concurrent `--worktree` processes, `kill -9` one — verify only the killed process's worktree is cleaned up

### DIFF
--- a/codeflash/optimization/optimizer.py
+++ b/codeflash/optimization/optimizer.py
@@ -22,6 +22,7 @@ from codeflash.code_utils.code_utils import cleanup_paths, get_run_tmp_file
 from codeflash.code_utils.env_utils import get_pr_number, is_pr_draft
 from codeflash.code_utils.git_utils import check_running_in_git_repo, git_root_dir
 from codeflash.code_utils.git_worktree_utils import (
+    cleanup_stale_worktrees,
     create_detached_worktree,
     create_diff_patch_from_worktree,
     create_worktree_snapshot_commit,
@@ -737,6 +738,8 @@ def mirror_path(path: Path, src_root: Path, dest_root: Path) -> Path:
 def run_with_args(args: Namespace) -> None:
     import atexit
     import signal
+
+    cleanup_stale_worktrees()
 
     optimizer = None
     original_sigterm = signal.getsignal(signal.SIGTERM)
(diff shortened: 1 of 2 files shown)


codeflash/optimization/optimizer.py
```diff
@@ -745,3 +772,9 @@ def run_with_args(args: Namespace) -> None:
             optimizer.cleanup_temporary_paths()
 
         raise SystemExit from None
+    finally:
+        atexit.unregister(cleanup_worktree_on_exit)
+        signal.signal(signal.SIGTERM, original_sigterm)
+        signal.signal(signal.SIGHUP, original_sighup)
+        signal.signal(signal.SIGQUIT, original_sigquit)
+        signal.signal(signal.SIGPIPE, original_sigpipe)
```
HeshamHM28, line 780: Q: Did we clean the worktree twice here, or am I misunderstanding this part?

PR #1836, opened 2026-03-16
fix: symlink node_modules in worktree for JS/TS projects

mohammedahmed18:
## Summary

- Creates a symlink from the worktree's `node_modules` to the original repo's `node_modules` when running in worktree mode for JS/TS projects, avoiding the need to reinstall dependencies
- Moves `mirror_path` utility from `optimizer.py` to `git_utils.py` for reuse across modules
- Updates `setup_test_config` signature across all `LanguageSupport` implementations (base, Python, Java, JavaScript) to accept `current_worktree` parameter

## Test plan

- [ ] Verify JS/TS optimization in worktree mode creates the symlink correctly
- [ ] Verify non-worktree mode (current_worktree=None) doesn't attempt symlink creation
- [ ] Verify Python and Java paths are unaffected by the signature change

### DIFF
--- a/tests/test_languages/test_javascript_setup_test_config.py
+++ b/tests/test_languages/test_javascript_setup_test_config.py
@@ -1,0 +1,149 @@
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+from codeflash.languages.javascript.support import JavaScriptSupport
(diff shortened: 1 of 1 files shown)


Saga4 (dismissed):
LGTM, add unit tests as well to this

PR #1848, opened 2026-03-17
fix: filter Java runtime annotations by class name and fix ordering

misrasaurabh1:
Runtime annotations in PR descriptions were broken in two ways:
1. add_runtime_comments() ignored class/method prefixes in keys, causing annotations from unrelated test classes to leak across files and sum incorrectly at the same line number. Now filters by class names found in each test source file.
2. Test functions were removed before annotations were added, shifting line numbers so annotations landed on wrong lines. Swapped ordering so annotations are applied first, then function removal carries them along correctly.

### DIFF
--- a/tests/test_languages/test_java/test_java_runtime_comments.py
+++ b/tests/test_languages/test_java/test_java_runtime_comments.py
@@ -135,6 +135,71 @@ public class FibTest {
         result = add_runtime_comments(source, original, optimized)
         assert result == source
 
+    def test_annotations_only_for_matching_classes(self) -> None:
+        """Bug: add_runtime_comments ignores class/method prefixes and only uses line numbers.
+
+        When runtime keys from different test classes map to the same line number,
+        annotations from unrelated classes leak into the source. Only annotations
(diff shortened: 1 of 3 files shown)


tests/test_languages/test_java/test_java_runtime_comments.py
```diff
@@ -135,6 +135,71 @@ def test_key_without_line_prefix_ignored(self) -> None:
...
+        # Only FibTest annotation should appear, NOT the summed value from both classes
+        expected = """\
+package com.example;
+
+import org.junit.jupiter.api.Test;
+
+public class FibTest {
+    @Test
+    void testSmall() {
+        Fibonacci.fibonacci(5); // 500\u03bcs -> 50.0\u03bcs (900% faster)
```
aseembits93, line 173: does `500\u03bcs` this look ok? is it rendered differently in java? @misrasaurabh1 @HeshamHM28 @mashraf-222

misrasaurabh1: i think it just is how microsecond is rendered (mu symbol)

PR #1855, opened 2026-03-17
fix: git diff auto-detection filters by current language instead of hardcoding .py

mashraf-222:
## Summary

- `get_git_diff()` in `code_utils/git_utils.py` hardcoded `.py` as the only valid file extension when parsing git diffs
- This caused the auto-detect flow (running codeflash without `--file`) to return **0 functions** for Java, JavaScript, and TypeScript projects
- This broke the Claude Code plugin integration where the stop hook runs `codeflash --subagent` without specifying `--file`, relying on git diff auto-detection to find changed functions

## Problems fixed

**Symptom:** When running `codeflash --subagent` (or any invocation without `--file`) on a Java/JS/TS project, the pipeline discovers no functions to optimize and exits silently — even when the git diff contains changed functions.

**Root cause:** Line 42 of `codeflash/code_utils/git_utils.py` had `if file_path.suffix != ".py": continue`, which filtered out all non-Python files from the git diff results. This function is called by both `get_functions_within_git_diff()` and `get_functions_within_commit()` in `discovery/functions_to_optimize.py`.

## Solution

Replaced the hardcoded `.py` check with `current_language_support().file_extensions`, which returns the correct extensions for whatever language is active in the session (e.g., `(".java",)` for Java, `(".js", ".jsx", ".mjs")` for JavaScript). The current language is already set by the time `get_git_diff()` is called during an optimization run.

## Code changes

- **`codeflash/code_utils/git_utils.py`**: Import `current_language_support`, resolve extensions once at function entry, use `not in supported_extensions` instead of `!= ".py"`
- **`tests/test_git_utils.py`**: Added 3 new tests:
  - `test_java_diff_found_when_language_is_java` — Java files discovered when language is Java
  - `test_java_diff_ignored_when_language_is_python` — Java files filtered out when language is Python
  - `test_mixed_lang_diff_filters_by_current_language` — mixed Python+Java diff correctly filters by active language

## Testing

```bash
$ uv run pytest tests/test_git_utils.py -v
# 15 passed (12 existing + 3 new) — no regressions
```

### DIFF
--- a/codeflash/benchmarking/plugin/plugin.py
+++ b/codeflash/benchmarking/plugin/plugin.py
@@ -267,9 +267,9 @@ class CodeFlashBenchmarkPlugin:
             os.environ["CODEFLASH_BENCHMARK_LINE_NUMBER"] = str(line_number)
             os.environ["CODEFLASH_BENCHMARKING"] = "True"
             # Run the function
-            start = time.time_ns()
+            start = time.perf_counter_ns()
             result = func(*args, **kwargs)
-            end = time.time_ns()
+            end = time.perf_counter_ns()
             # Reset the environment variable
             os.environ["CODEFLASH_BENCHMARKING"] = "False"
 


aseembits93 (dismissed):

aseembits93:
@claude fix the failing windows unit test

PR #1859, opened 2026-03-18
feat: multi-language orchestration loop with per-language config discovery

mashraf-222:
## Summary

Adds multi-language orchestration to the Codeflash CLI. The optimizer now discovers all language configs in a project (Python, Java, JS/TS), and runs a full optimization pass for each language automatically.

**This PR is part of a 3-repo change set:**
- **codeflash** (this PR) — Multi-language orchestration loop, config discovery, and language-agnostic git diff
- **codeflash-cc-plugin** — Unified hook that triggers a single `codeflash --subagent` call
- **optimize-me** — Mixed-language test fixture (Python + Java + JS/TS) for E2E validation

The CLI orchestration loop is the core — the cc-plugin delegates to it, and optimize-me validates it end-to-end.

### Config Discovery
- `find_all_config_files()` walks CWD→root, discovers `pyproject.toml` (Python), `codeflash.toml` (Java), `package.json` (JS/TS). Closest config wins per language.
- New `LanguageConfig` dataclass holds config dict, path, and language enum per discovered config.
- Extracted `normalize_toml_config()` shared helper for consistent config normalization.

### Orchestration Loop
- `main()` iterates over discovered configs, deep-copies args per language, calls `apply_language_config()` then `optimizer.run_with_args()` for each.
- `--file` flag filters to the matching language only. `--all` and no-flags run all discovered languages.
- Per-language error isolation — one language failing doesn't block others.
- Summary logging after all passes complete.

### Auto-Detection
- `detect_unconfigured_languages()` compares configs vs git diff to find languages with changed files but no config.
- `auto_configure_language()` creates on-the-fly configs by detecting project roots (pom.xml, package.json).

### Language-Agnostic Git Diff
- `get_git_diff()` now uses `get_supported_extensions()` from the registry instead of the singleton. The coarse filter lets through all supported files; per-file language detection happens downstream.

## Tests

- 73 tests in `test_multi_language_orchestration.py` (orchestration loop, apply_language_config, summary logging, CLI path routing, unconfigured detection)
- 11 tests in `test_multi_config_discovery.py` (find_all_config_files, LanguageConfig)
- 15 tests in `test_git_utils.py` (language-agnostic diff filtering)
- Registry test pollution fix in `test_registry.py`

All tests pass locally.

## E2E Validation

Full E2E session with local backend services — Fibonacci (java-test-project), encodedLength (aerospike), isPalindrome + mergeSorted (optimize-me Java), retryWithBackoff (optimize-me JS) all passed.

### [codeflash-cc-plugin — Unified multi-language hook](https://github.com/codeflash-ai/codeflash-cc-plugin/pull/34)
The cc-plugin is the entry point that triggers Codeflash from Claude Code. Previously it had 3 separate per-language code paths that each invoked the CLI differently. With the orchestration loop introduced in this PR, the cc-plugin was simplified to a single `codeflash --subagent` call — it only needs to detect that *any* config exists and find the binary. All multi-language discovery and dispatch is now handled by `find_all_config_files()` and the orchestration loop in `main.py` from this PR.

### [optimize-me — Mixed-language test fixture](https://github.com/codeflash-ai/optimize-me/pull/305)
The optimize-me repo was extended with Java and JS/TS subprojects to serve as an E2E validation target for the orchestration loop. Each subproject has its own codeflash config file, so running Codeflash at the repo root exercises `find_all_config_files()` discovering multiple configs and the loop running separate optimization passes per language. The E2E validation results above were produced by running this PR's code against the optimize-me fixture.

### DIFF
--- a/codeflash/cli_cmds/cli.py
+++ b/codeflash/cli_cmds/cli.py
@@ -328,6 +328,11 @@ def apply_language_config(args: Namespace, lang_config: LanguageConfig) -> Names
         args.benchmarks_root = Path(args.benchmarks_root).resolve()
     args.test_project_root = project_root_from_module_root(args.tests_root, config_path)
 
+    if is_java and config_path.is_dir():
+        # For Java projects, config_path IS the project root directory (from build-tool detection).
+        args.project_root = config_path.resolve()
+        args.test_project_root = config_path.resolve()
+
     return args
 
 
--- a/codeflash/code_utils/config_parser.py
+++ b/codeflash/code_utils/config_parser.py
@@ -157,6 +157,66 @@ def normalize_toml_config(config: dict[str, Any], config_file_path: Path) -> dic
     return config
 
 
+def _parse_java_config_for_dir(dir_path: Path) -> dict[str, Any] | None:
+    from codeflash.languages.java.build_tools import parse_java_project_config
+
+    return parse_java_project_config(dir_path)
+
+
+_SUBDIR_SKIP = frozenset({
+    ".git", ".hg", ".svn", "node_modules", ".venv", "venv", "__pycache__",
+    "target", "build", "dist", ".tox", ".mypy_cache", ".ruff_cache", ".pytest_cache",
+})
+
+
+def _check_dir_for_configs(
+    dir_path: Path,
+    configs: list[LanguageConfig],
+    seen_languages: set[Language],
+) -> None:
+    """Check a single directory for language config files and append any found to *configs*."""
+    if Language.PYTHON not in seen_languages:
+        pyproject = dir_path / "pyproject.toml"
+        if pyproject.exists():
+            try:
+                with pyproject.open("rb") as f:
+                    data = tomlkit.parse(f.read())
+                tool = data.get("tool", {})
+                if isinstance(tool, dict) and "codeflash" in tool:
+                    raw_config = dict(tool["codeflash"])
+                    normalized = normalize_toml_config(raw_config, pyproject)
+                    seen_languages.add(Language.PYTHON)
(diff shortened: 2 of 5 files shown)


codeflash/cli_cmds/cli.py
```diff
@@ -108,10 +109,18 @@ def process_pyproject_config(args: Namespace) -> Namespace:
     assert args.module_root is not None, "--module-root must be specified"
     assert Path(args.module_root).is_dir(), f"--module-root {args.module_root} must be a valid directory"
 
-    # For JS/TS projects, tests_root is optional (Jest auto-discovers tests)
-    # Default to module_root if not specified
-    is_js_ts_project = pyproject_config.get("language") in ("javascript", "typescript")
-    is_java_project = pyproject_config.get("language") == "java"
+    # Determine project language from explicit config or config file name.
+    # Java projects use codeflash.toml but don't always have an explicit language field.
+    config_language = pyproject_config.get("language")
```
misrasaurabh1: why would we store language config in pyproject.toml file which is python exclusive? this design is incorrect

mashraf-222: updated it

code_to_optimize/java/src/main/java/com/example/Fibonacci.java
```diff
@@ -19,7 +19,7 @@ public static long fibonacci(int n) {
             throw new IllegalArgumentException("Fibonacci not defined for negative numbers");
         }
         if (n <= 1) {
-            return n;
+            return n; // base case
```
mashraf-222: I am reverting these changes also

PR #1860, opened 2026-03-18
fix: instrument attrs classes in __init__ capture (crash fix + monkey-patch wrapper)

KRRT7:
## Problem

When codeflash tried to optimize a method on an `@attrs.define` (or `@attr.s`) class, it crashed with:

```
TypeError: super(type, obj): obj (instance of X) is not an instance or subtype of X
```

**Root cause:** `attrs.define(slots=True)` (the default) replaces the original class with a brand-new slots class at import time. The synthetic `def __init__(self, *args, **kwargs): super().__init__(...)` that codeflash injects into the class body bakes a `__class__` cell pointing to the *original* class, but `self` is already an instance of the *new* slots class — so the `super()` call explodes.

Additionally, the test-generation context extractor had no knowledge of attrs field conventions, so it couldn't build correct `__init__` stubs for attrs classes.

### 1. Crash fix → module-level monkey-patch wrapper (`instrument_codeflash_capture.py`)

Instead of injecting a synthetic `__init__` body into the class (which triggers the crash), we now emit a **module-level patch block** immediately after the class definition:

```python
_codeflash_orig_MyClass_init = MyClass.__init__
def _codeflash_patched_MyClass_init(self, *args, **kwargs):
    return _codeflash_orig_MyClass_init(self, *args, **kwargs)
MyClass.__init__ = codeflash_capture(...)(_codeflash_patched_MyClass_init)
```

The wrapper is a plain module-level function with no `__class__` cell, so it's immune to the slot-class replacement. Covers `@attrs.define`, `@attrs.define(frozen=True)`, `@attrs.mutable`, `@attrs.frozen`, `@attr.s`, `@attr.attrs`.

### 2. Test-gen context: attrs support (`code_context_extractor.py`)

- Added `_get_attrs_config()` helper (parallel to `_get_dataclass_config`)
- `_collect_synthetic_constructor_type_names`: attrs classes now included
- `_build_synthetic_init_stub`: generates correct stub with `kw_only` support
- `_extract_synthetic_init_parameters`: handles attrs `factory=` keyword (equivalent to dataclass `default_factory=`)

### 3. Tests

6 new tests, all with exact expected-output assertions:
- 3 in `test_instrument_codeflash_capture.py` — verify the monkey-patch block is emitted correctly for `@attrs.define`, `@attrs.define(frozen=True)`, and `@attr.s`
- 3 in `test_code_context_extractor.py` — verify `__init__` stub generation for attrs classes (required fields, `factory=` defaults, `init=False`)

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/codeflash/code_utils/git_utils.py
+++ b/codeflash/code_utils/git_utils.py
@@ -13,7 +13,6 @@ from rich.prompt import Confirm
 from unidiff import PatchSet
 
 from codeflash.cli_cmds.console import logger
-from codeflash.languages.registry import get_supported_extensions
 
 if TYPE_CHECKING:
     from git import Repo
@@ -39,7 +38,6 @@ def get_git_diff(
         uni_diff_text = repository.git.diff(
             commit.hexsha + "^1", commit.hexsha, ignore_blank_lines=True, ignore_space_at_eol=True
         )
-    supported_extensions = set(get_supported_extensions())
     patch_set = PatchSet(StringIO(uni_diff_text))
     change_list: dict[str, list[int]] = {}  # list of changes
     for patched_file in patch_set:


KRRT7:
@claude FAILED tests/test_git_utils.py::TestGetGitDiffMultiLanguage::test_java_diff_ignored_when_language_is_python - AssertionError: assert 1 == 0 fix this

KRRT7:
@claude fix the root cause of the failing tests

PR #1880, opened 2026-03-20
feat: zero-config Java projects + smart ReplayHelper for end-to-end optimization

misrasaurabh1:
## Summary

Eliminates `codeflash.toml` for Java projects and fixes the complete trace → optimize pipeline to work end-to-end on real Java projects (validated on aerospike-client-java).

### Zero-config Java support
- Auto-detect Java projects from `pom.xml` / `build.gradle` — no config file needed
- Read custom settings from `pom.xml` `<properties>` or `gradle.properties` (`codeflash.*` keys)
- Multi-module Maven scanning: parses each module's `<sourceDirectory>` / `<testSourceDirectory>`, picks module with most Java files as source root
- Deleted all `codeflash.toml` files

### Smart ReplayHelper (behavior + performance parity)
- `ReplayHelper.replay()` now reads `CODEFLASH_MODE` env var and produces the same output as existing test instrumentation
- Behavior mode: captures return value via Kryo, writes to SQLite `test_results` table for correctness comparison
- Performance mode: runs inner loop for JIT warmup, prints timing markers matching the optimizer's expected format
- No mode: just invokes (trace-only or manual testing)

### Bug fixes
- **JFR parser**: normalize `/` → `.` in class names (JVM internal format vs Java package format)
- **Graceful timeout**: send SIGTERM before SIGKILL so JFR can dump recording and shutdown hooks run
- **TracingTransformer**: remove `isRecording()` check that prevented instrumenting classes loaded during serialization (was causing 3 captures instead of 10,000+)
- **Replay test generator**: JUnit 4 support (`org.junit.Test` vs `org.junit.jupiter.api.Test`), detect from project build config
- **Overloaded methods**: global counter per method name to avoid duplicate replay test method names
- **Instrumentation**: fix `_add_behavior_instrumentation` for compact `@Test` lines (annotation + signature on same line)
- **project_root**: use build root directory (not sub-module) for multi-module Maven projects
- **optimize subparser**: `add_help=False` so `-h` in Java commands isn't intercepted as `--help`

### Validated end-to-end on aerospike-client-java
- 10,500+ invocations traced across 282 methods
- 41 functions ranked by JFR CPU profiling data
- 55 replay test files generated (JUnit 4 compatible)
- Replay tests compile, run, and pass (129 tests for `Crypto.computeDigest`)
- Behavior baseline established with timing data (4.81ms over 119 loops)
- Candidates correctly verified and rejected when behavior doesn't match

## Test plan

- [x] 33 config detection tests (build tool, source/test root, Maven/Gradle properties, multi-module)
- [x] 13 JFR parser tests (normalization, filtering, ranking, timeout, project_root)
- [x] 10 replay test generation tests (JUnit 4/5, overloads, instrumentation)
- [x] 8 tracer e2e tests (agent capture, replay generation, orchestration)
- [x] 6 integration tests (full pipeline: discover → rank → compile)
- [x] 2 replay test discovery tests
- [x] Full optimizer pipeline on aerospike benchmark: trace → discover → rank → optimize → verify

### DIFF
--- a/codeflash/languages/base.py
+++ b/codeflash/languages/base.py
@@ -50,6 +50,12 @@ class IndexResult:
     error: bool
 
 
+@dataclass(frozen=True)
+class SetupError:
+    message: str
+    should_abort: bool
+
+
 @dataclass
 class HelperFunction:
     """A helper function that is a dependency of the target function.
@@ -725,8 +731,12 @@ class LanguageSupport(Protocol):
         """Parse/validate a module before optimization."""
         ...
 
-    def setup_test_config(self, test_cfg: TestConfig, file_path: Path, current_worktree: Path | None = None) -> None:
-        """One-time project setup after language detection. Default: no-op."""
+    def setup_test_config(self, test_cfg: TestConfig, file_path: Path, current_worktree: Path | None = None) -> bool:
+        """One-time project setup after language detection. Default: no-op.
+
+        Returns True if the project is valid for optimization, False otherwise.
+        """
+        return True
 
     def adjust_test_config_for_discovery(self, test_cfg: TestConfig) -> None:
         """Adjust test config before test discovery. Default: no-op."""
--- a/codeflash/languages/java/build_tool_strategy.py
+++ b/codeflash/languages/java/build_tool_strategy.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import logging
 import os
+import shutil
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
@@ -72,6 +73,18 @@ class BuildToolStrategy(ABC):
             return dev_jar_gradle
 
         return None
+
+    def find_wrapper_executable(
+        self, build_root: Path, wrapper_names: tuple[str, ...], system_command: str
+    ) -> str | None:
+        search = build_root.resolve()
+        while search != search.parent:
+            for name in wrapper_names:
+                candidate = search / name
+                if candidate.exists():
+                    return str(candidate)
+            search = search.parent
+        return shutil.which(system_command)
 
     @abstractmethod
     def find_executable(self, build_root: Path) -> str | None:
--- a/codeflash/languages/java/gradle_strategy.py
+++ b/codeflash/languages/java/gradle_strategy.py
@@ -417,22 +417,7 @@ class GradleStrategy(BuildToolStrategy):
         )
 
     def find_executable(self, build_root: Path) -> str | None:
-        # Walk up from build_root to find gradlew — for multi-module projects
-        # the wrapper lives at the repo root, which may be a parent of build_root.
-        current = build_root.resolve()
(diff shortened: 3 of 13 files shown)


mashraf-222:
## Review & fix: pom.xml formatting preservation (3c63b60)

Reviewed the PR changes and identified one genuine issue in the config writer.

### What was fixed

**`_write_maven_properties()` and `_remove_java_build_config()` destroy pom.xml formatting** — both used `xml.etree.ElementTree` which strips XML comments, mangles namespace declarations (`xmlns` → `ns0:` prefixes), and reformats all whitespace. Replaced with text-based regex manipulation that only touches `codeflash.*` property lines and preserves everything else. The new approach also detects existing indentation from sibling elements so inserted properties match the file's style.

Additionally:
- Extracted duplicated `key_map` dict to shared `_MAVEN_KEY_MAP` module constant
- Aligned `_remove_java_build_config()` to check pom.xml first, matching write priority
- Added 9 tests (7 for config_writer, 2 for detector confirming zero-config intent)

### What was reviewed and left as-is

Confirmed the following are intentional design decisions, not bugs:
- **`has_existing_config()` returning True for any Java build file** — this is the zero-config mechanism that skips the first-run setup flow
- **`isRecording()` guard removal in TracingTransformer** — PR description explains this fixed captures going from 3 to 10,000+
- **`add_help=False` on optimize subparser** — prevents argparse from intercepting `-h` meant for Java command passthrough

### Validation

- All 152 setup tests pass (including 9 new)
- All 8 tracer E2E tests pass
- Full Fibonacci E2E optimization: 279x speedup found, mark-as-success 200
- prek clean (ruff check + format)

mashraf-222:
3 commits addressing the remaining open TODOs from the code review:

### 1. `5942ae9e` — Monorepo config priority 
`parse_config_file()` used to short-circuit on the first Java build file found, before the existing depth-comparison logic for package.json/pyproject.toml. In a monorepo with a parent `pom.xml` and a child `package.json`, Java would incorrectly win. Now all config sources are detected first and the closest one to CWD wins.

### 2. `12921447` — Tracer line numbers + dropped captures 
- **Line numbers**: `TracingClassVisitor` hardcoded `0` because ASM's `visitMethod()` doesn't provide line info. Added a pre-scan pass in `instrumentClass()` that collects first line numbers via `visitLineNumber()` before the instrumentation pass.
- **Dropped captures**: Serialization timeouts silently dropped invocations with no tracking. Added `AtomicInteger droppedCaptures` counter and included it in flush metadata.

### 3. `970c9f86` — Package detection scope 
Changed `detect_packages_from_source()` from 2 package components to 3, so `com.aerospike.client.util` → `com.aerospike.client` (was `com.aerospike`). Reduces instrumentation blast radius from entire organization to actual source package.

### E2E validation
Fibonacci optimization: 258x speedup. Full pipeline (discovery → test gen → instrumentation → benchmarking → candidate evaluation) passed.

PR #1950, opened 2026-04-01
fix: reject optimizations that modify class members without changing target method

mashraf-222:
## Problem

The optimizer accepts LLM output that modifies constructors, adds fields, or adds helper methods without actually changing the target method. This causes 15+ wrong-method PRs where the PR title says one method but the diff changes constructors or adds unused members.

Examples from Zuul/Eureka:
- `Header.getValue` — title says getValue, diff changes constructor
- `FilterType.toString` — diff removes private keyword, doesn't touch toString
- `MutableFilterRegistry.size` — changes from delegating to cached count
- `SessionContext.getRouteVIP` — correctness bug, returns wrong value
- `EurekaHttpResponse.getStatusCode` — title says getStatusCode, diff changes constructor

## Root Cause

`replace_function()` in `replacement.py` blindly applies all class member changes from the LLM output (constructors via `_replace_constructors`, fields and helpers via `_insert_class_members`) without checking whether the target method itself was modified.

## Fix

Added a guard in `replace_function()` that:
1. Detects if the LLM produced any class-level modifications (fields, helpers, constructors)
2. Extracts the original target method source from the file
3. Compares it (normalized for whitespace) with the optimization's target method
4. If the target is unchanged → rejects the optimization as a no-op

Two helper functions added:
- `_normalized_equal()` — whitespace-insensitive comparison
- `_extract_original_method_source()` — extracts the original method from source (handles overloads via line numbers)

## Test Coverage

5 new tests in `TestUnchangedTargetWithClassModifications`:
- `test_unchanged_target_with_unused_field_rejected` — field added, target unchanged → rejected
- `test_unchanged_target_with_modified_constructor_rejected` — constructor modified, target unchanged → rejected (reproduces Zuul #52 pattern)
- `test_unchanged_target_with_helper_rejected` — helper added, target unchanged → rejected
- `test_changed_target_with_field_accepted` — target changed + field added → accepted (valid memoization)
- `test_changed_target_only_accepted` — target changed, no class modifications → accepted

All 37 tests pass (32 existing + 5 new).

Closes CF-1084


KRRT7:
@claude I believe the inner normalize function exists elsewhere in the codebase, double check

mashraf-222:
## Review

**Bug premise verified — real.** `replace_function` in `codeflash/languages/java/replacement.py` on main applies `parsed.new_fields` / `parsed.modified_constructors` / helpers without ever comparing the target-method text before and after parsing. That explains the Zuul/Eureka "wrong-method" PR class where the target method body is unchanged but constructors or class-level fields are modified.

**Fix is architecturally correct.** The guard at `replace_function` runs before `_insert_class_members` mutates source, follows the existing "return source → reject" pattern, and correctly distinguishes changes that modify the target body (accepted) from changes that don't (rejected).

**Relationship to PR #1947 (CF-1081):** overlapping but distinct.
- **#1947** rejects "additions exist AND target unchanged AND no addition identifier referenced in target body" — catches the commons-lang `NULL_SUPPLIER` pattern.
- **#1950** (this PR) rejects "additions OR modified constructors exist AND target unchanged" — catches the Zuul `Header.getValue` constructor-precomputation pattern.

Neither is strictly a subset. Recommend merging both, with #1947 merged first since its check is narrower; then this PR's broader guard covers the constructor case #1947 skips. A follow-up consolidation (single helper + one `normalize_java_code` via `formatter.py:264` instead of the inlined duplicate) is worth doing after both land.

1. `prek / prek` was failing because the prek `mypy` hook ran on `tests/test_languages/test_java/test_replacement.py` and flagged every untyped test — pre-existing pattern that the hook surfaced because this PR touches the file. Added `-> None` + parameter type annotations on all 37 test methods and the `java_support` fixture.

Snyk "code test limit reached" is an account quota, unrelated. Ready for re-review.

PR #1999, opened 2026-04-04
[FIX][JS] Avoid adding .js to relative imports in tests with jest as it's using its own resolver

mohammedahmed18:


mohammedahmed18:
closing, it's resolved in https://github.com/codeflash-ai/codeflash/pull/2020

PR #2062, opened 2026-04-10
fix: share CLI flags between main parser and optimize subparser

mashraf-222:
## Problem

30 CLI flags (`--verbose`, `--no-pr`, `--file`, `--function`, `--all`, `--trace-only`, etc.) were defined only on the main `ArgumentParser`. When using the `optimize` subparser path (`codeflash optimize mvn test ...`), these flags were silently dropped by argparse — the subparser didn't recognize them and they were treated as part of the Java command.

This broke virtually every flag for the `optimize` subparser path, which is the primary entry point for Java optimization.

## Root Cause

`_build_parser()` in `cli.py` added all 30+ flags to the main parser only. The `optimize` subparser was created with `subparsers.add_parser("optimize")` but didn't inherit any of these flags.

## Fix

Used argparse's `parents=[]` pattern:
1. Created `shared_flags = ArgumentParser(add_help=False)` containing all shared flags
2. Both the main parser and optimize subparser use `parents=[shared_flags]`
3. Added `--trace-only` flag (was missing entirely)
4. Aligned `max_function_count` default to 256 (was 100 on subparser vs 256 elsewhere)

## Test Coverage

- 10 new unit tests in `tests/test_cli_args.py`:
  - 7 tests verify flags work on the optimize subparser (`--verbose`, `--no-pr`, `--file`, `--function`, multiple flags, command isolation, max_function_count default)
  - 3 tests verify flags still work on the main parser path (no regression)

## Testing

```
$ uv run pytest tests/test_cli_args.py -v
10 passed
```

### DIFF
--- a/codeflash/cli_cmds/cli.py
+++ b/codeflash/cli_cmds/cli.py
@@ -374,14 +374,115 @@ def _handle_reset_config(confirm: bool = True) -> None:
 
 @lru_cache(maxsize=1)
 def _build_parser() -> ArgumentParser:
-    parser = ArgumentParser()
+    # Shared flags used by both the main parser and the optimize subparser.
+    # Using argparse parents= ensures flags like --verbose, --no-pr, --file are recognized
+    # regardless of whether the user runs "codeflash --verbose" or "codeflash optimize --verbose".
+    shared_flags = ArgumentParser(add_help=False)
+    shared_flags.add_argument("--file", help="Try to optimize only this file")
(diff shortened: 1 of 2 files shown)


KRRT7:
Hey @mashraf-222 — a few things on your open PRs:

- **#2066** has a review comment from @HeshamHM28 — please address it.
- Make sure each PR has unit tests covering the fix. #2062 and #2066 look OK, #2063 could use more coverage.
- Rebase all three on main — CI has changed recently.

PR #2066, opened 2026-04-10
fix: reject JDK < 11 early with clear error message

mashraf-222:
## Problem

When running codeflash with JDK 8, the pipeline fails late with a cryptic `UnsupportedClassVersionError` from the runtime JAR (class version 55.0 requires JDK 11+). The error gives no guidance about what version is needed.

## Root Cause

`ensure_runtime_environment()` in `support.py` detects the Java version via `_detect_java_version()` but never validates it meets the minimum requirement (JDK 11). The version is stored for API payloads but not checked.

## Fix

Added a version check immediately after `_detect_java_version()`: if major version < 11, logs a clear error message explaining that JDK 11+ is required and returns False. This stops the pipeline before it hits the cryptic class version error.

## Test Coverage

3 new tests in `TestJdkVersionCheck`:
- `test_jdk8_rejected`: JDK 8 returns False
- `test_jdk11_accepted`: JDK 11 returns True
- `test_jdk21_accepted`: JDK 21 returns True

## Testing

```
$ uv run pytest tests/test_languages/test_java/test_support.py::TestJdkVersionCheck -v
3 passed
```

### DIFF
--- a/codeflash/languages/java/support.py
+++ b/codeflash/languages/java/support.py
@@ -583,9 +583,9 @@ class JavaSupport(LanguageSupport):
                     end = line.find('"', start + 1)
                     if start != -1 and end != -1:
                         full_version = line[start + 1 : end]
-                        # Use major version only: "17.0.2" -> "17", "1.8.0_292" -> "8"
-                        major = full_version.split(".")[0]
-                        self._language_version = "8" if major == "1" else major
+                        # Use major version only: "17.0.2" -> "17". JDK 8 and earlier (reported as
+                        # "1.x.y") are unsupported — the downstream minimum-version check rejects them.
+                        self._language_version = full_version.split(".")[0]
                         return
         except Exception:
             pass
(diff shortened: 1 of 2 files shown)


codeflash/languages/java/support.py
```diff
@@ -546,6 +546,21 @@ def ensure_runtime_environment(self, project_root: Path) -> bool:
...
+                    logger.error(
+                        "Java %s detected, but codeflash requires JDK 11 or later. "
+                        "The codeflash-runtime JAR and --add-opens flags are incompatible with JDK %s. "
+                        "Please install JDK 11+ and ensure it is on your PATH.",
+                        self._language_version,
+                        self._language_version,
+                    )
+                    return False
+            except ValueError:
+                pass
```
HeshamHM28, line 562: Set the minimum Java version and remove the --add-opens options for Java 8.

Issue #526, opened 2025-07-07
Refactor Large Functions


Saga4:
10 Large Functions which could be needing refactoring: 

- [ ]   1. create_wrapper_function - codeflash/code_utils/instrument_existing_tests.py
- [ ] 
- [ ]   2. collect_setup_info - codeflash/cli_cmds/cmd_init.py
- [ ] 
- [ ]   3. comparator - codeflash/verification/comparator.py
- [ ] 
- [ ]   4. determine_best_candidate - codeflash/optimization/function_optimizer.py
- [ ] 
- [ ]   5. parse_test_xml - codeflash/verification/parse_test_output.py
- [ ] 
- [ ]   6. tabulate - codeflash/code_utils/tabulate.py
- [ ] 
- [ ]   7. process_test_files - codeflash/discovery/discover_unit_tests.py
- [ ] 
- [ ]   8. install_github_actions - codeflash/cli_cmds/cmd_init.py
- [ ] 
- [ ]   9. print_stats - codeflash/tracer.py
- [ ] 
- [ ]   10. establish_original_code_baseline - codeflash/optimization/function_optimizer.py

shreejaykurhade:
I would like to take up this. I'll start with 

1. create_wrapper_function - codeflash/code_utils/instrument_existing_tests.py

/assign

PR #2072, opened 2026-04-20
Refactor: comparator to extract shared string, sequence, and mapping comparison helpers

shreejaykurhade:
This PR refactors codeflash/verification/comparator.py to reduce duplication inside comparator without changing its special-case dispatch flow. #526

## The existing function had several repeated comparison bodies across:

temp-path-aware string handling
ordered recursive comparisons for lists/tuples and similar containers
mapping/superset comparison logic for dict-like objects

This change extracts those repeated bodies into small helpers while keeping the main if/elif structure, branch ordering, and existing inline comments intact.

## What Changed

Added three focused private helpers:

_compare_string_values
_compare_ordered_values
_compare_mapping_values

## Updated comparator to reuse those helpers in the existing branches for:

str and string subclasses
list, tuple, array.array, and NumbaList
dict, the later dict-like fallback, and NumbaDict

## Why This Is Better

This improves maintainability by:

removing duplicated recursive comparison logic
centralizing shared comparison behavior in one place
reducing the chance of fixing one branch and missing another
preserving the readability of the main comparator flow instead of turning it into a large rewrite

## Behavior / Compatibility

This PR is intended to be behavior-preserving.

## Notably:

special-case branches were not removed
dispatch order inside comparator was preserved
existing inline comments in the main function were kept
this is a targeted refactor, not a redesign of comparator behavior

## Validation

uv run pytest -q tests/test_comparator.py
uv run ruff check codeflash/verification/comparator.py

## Results:

160 passed, 43 skipped
ruff passed
image

## Risk

Low.

The change is limited to extracting repeated logic into helpers and is covered by the dedicated comparator test suite.

got this on uv run pytest. unrelated errors.
image

### DIFF
--- a/codeflash/verification/comparator.py
+++ b/codeflash/verification/comparator.py
@@ -203,7 +203,13 @@ def _compare_mapping_values(orig: Any, new: Any, superset_obj: bool) -> bool:
 
 
 def comparator(orig: Any, new: Any, superset_obj: bool = False) -> bool:
-    """Compare two objects for equality recursively. If superset_obj is True, the new object is allowed to have more keys than the original object. However, the existing keys/values must be equivalent."""
+    """Recursively compare captured values while preserving Codeflash-specific equivalence rules.
+
+    This comparator handles plain Python containers, framework objects, wrapped exceptions,
+    and a handful of normalized cases such as temporary file paths. When ``superset_obj`` is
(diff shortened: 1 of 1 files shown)


shreejaykurhade:
@Saga4 and @KRRT7 please check it

PR #2075, opened 2026-04-20
Refactor: Java wrapper call instrumentation into focused helpers

shreejaykurhade:
This PR refactors wrap_target_calls_with_treesitter in codeflash/languages/java/instrumentation.py to reduce complexity without changing instrumentation behavior. Issue #526

## The original function handled several responsibilities in one place:

- filtering and ordering target calls
- precomputing byte/character offsets for replacement
- building timing/capture statements for instrumented calls
- generating replacement code for expression-statement calls
- generating prefix blocks for embedded calls
- applying back-to-front source rewrites

That made the function harder to review and riskier to modify, especially in a part of the codebase with a lot of exact emitted-source expectations.

## What Changed

The repeated preparation and code-generation logic is now split into focused private helpers:

- _prepare_valid_calls
- _build_call_statements
- _build_expression_statement_replacement
- _build_embedded_call_prefix_lines

## wrap_target_calls_with_treesitter now acts as the orchestration layer for:

- parsing the body with tree-sitter
- collecting candidate calls
- preparing valid call metadata
- iterating through calls in reverse order
- applying expression or embedded-call replacements

## Why This Is Better

This improves maintainability by:

- reducing the responsibility surface of the main wrapper function
- separating metadata preparation from emitted-code construction
- making the expression-statement and embedded-call paths easier to reason about in isolation
- lowering the risk of future edits to one instrumentation path affecting another

This is intended to be a behavior-preserving refactor, not a logic change.

## Behavior / Compatibility

No instrumentation behavior is intended to change.

## In particular, this refactor preserves:

- lambda and complex-expression filtering
- call ordering and reverse-application semantics
- void vs non-void handling
- precise-timing vs non-precise-timing behavior
- emitted marker / serialization / SQLite-write structure

## Validation
```bash
uv run pytest -q tests/test_languages/test_java/test_instrumentation.py
uv run ruff check codeflash/languages/java/instrumentation.py
```

## Results:

Both Passed
image
image

## Risk

Low.

This change is a structural refactor in an area already covered by exact-string Java instrumentation tests, and the refactor preserves the original control flow rather than redesigning the instrumentation logic.

### DIFF
--- a/codeflash/languages/java/instrumentation.py
+++ b/codeflash/languages/java/instrumentation.py
@@ -262,6 +262,206 @@ def _generate_sqlite_write_code(
     ]
 
 
+def _prepare_valid_calls(
+    calls: list[dict[str, Any]],
+    body_bytes: bytes,
+    body_lines: list[str],
+    line_byte_starts: list[int],
+) -> list[dict[str, Any]]:
+    """Filter collected calls and pre-compute offsets used during replacement."""
+    valid_calls = [call for call in calls if not call["in_lambda"] and not call.get("in_complex", False)]
+    if not valid_calls:
+        return []
+
+    valid_calls.sort(key=lambda call: call["start_byte"])
+    for counter, call in enumerate(valid_calls, 1):
+        call["_counter"] = counter
+        call["_call_start_char"] = len(body_bytes[: call["start_byte"]].decode("utf8"))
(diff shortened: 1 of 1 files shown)


shreejaykurhade:
@Saga4 @KRRT7 please review.

KRRT7:
Closing in favor of test coverage work — see #2079 for issues to pick up. Thanks for the contribution!

PR #2132, opened 2026-05-07
perf: targeted performance improvements for E2E pipeline

KRRT7:
Targeted performance improvements for the E2E optimization pipeline hot path, plus bug fixes discovered during profiling.

## PR stack (deep linear — merge bottom-up)

| # | PR | Branch | Description |
|---|---|---|---|
| 1 | #2145 | `fix/mypy-strict-errors` | fix: resolve mypy strict errors in models.py and code_utils.py |
| 2 | #2144 | `fix/test-files-silent-dedup` | fix: silently skip duplicate `TestFiles.add()` instead of raising |
| 3 | #2138 | `fix/windows-junitxml-path` | fix: Windows junitxml path handling |
| 4 | #2141 | `fix/subprocess-xml-errors` | fix: surface subprocess errors when pytest XML is missing |
| 5 | #2134 | `perf/discovery-prefilter` | perf: prefilter files by path before `read_text()` |
| 6 | #2135 | `perf/discovery-ast-cache` | perf: cache file reads and AST parses per discovery run |
| 7 | #2136 | `perf/discovery-js-read` | perf: read JS/TS files once during discovery export checks |
| 8 | #2137 | `cf-2132-skip-callers-json-decode` | perf: skip unused callers JSON decode in ProfileStats |
| 9 | #2143 | `perf/misc-hot-path` | perf: misc hot-path optimizations (lru_cache, copy, set dedup, .get()) |
| 10 | **#2132** | `cf-perf-improvements` | **this umbrella** (accumulates all above) |

Previously merged (already on main): #2128, #2129, #2130, #2131, #2140, #2142

## Design decisions

- **No behavioral changes** in perf PRs: every optimization preserves identical semantics
- **Bug fixes first**: fixes land before perf changes so they propagate through the stack
- **Deep linear stack**: each PR targets the branch below it; when bottom merges to main, GitHub auto-retargets the next one
- **Discovery-scoped cache**: `discovery_cache()` context manager ensures no cross-run leakage
- **Dead code removal**: `_find_all_functions_via_language_support()` removed (never called, had pre-existing type error)

## Test plan

- [x] `uv run prek` passes on all 9 layers individually
- [x] Full unit test suite passes (4206 tests, 0 regressions)
- [x] Java tracer/profiler e2e tests pass
- [x] JS/TS tests pass
- [ ] Cumulative benchmark (TBD)

PR #2133, opened 2026-05-07
fix: silently skip duplicate TestFiles.add() instead of raising

KRRT7:
## Parent PR

#2132 — perf: targeted performance improvements for E2E pipeline hot path

## Summary

Changes `TestFiles.add()` duplicate handling from raising `ValueError` to silently skipping (first-write-wins). Fixes Java e2e test failures where test discovery adds the same `instrumented_behavior_file_path` with different metadata fields.

## Problem

The set-based dedup introduced in #2130 correctly identifies duplicate test files by path. However, it raised `ValueError` on duplicates — which breaks Java test discovery where the same path appears multiple times with varying metadata.

Since `get_test_type_by_instrumented_file_path()` returns on first path match anyway, duplicates by path are dead weight. Raising is both incorrect (existing callers never expected it) and unnecessary.

## Changes

- `codeflash/models/models.py`: Remove `ValueError` raise in `add()`, silently return when path already in `_seen_paths`
- `tests/test_test_files_add.py`: Update test from `test_add_duplicate_raises` → `test_add_duplicate_is_noop`

## Key design decisions

- **First-write-wins**: The first `TestFile` added for a given path is kept; subsequent adds with the same path are silently discarded. This matches the existing lookup semantics.
- **No behavioral change for unique paths**: Only the duplicate path changes — unique additions work identically.

## Test plan

- [x] `test_add_duplicate_is_noop` verifies silent skip on duplicate path
- [x] `test_add_unique_test_file` passes
- [x] `test_add_many_files_performance` passes (100 files, O(1) per add)
- [ ] Java e2e tests pass (was failing before this fix)

### DIFF
--- a/codeflash/languages/python/parse_xml.py
+++ b/codeflash/languages/python/parse_xml.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import os
 import re
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from junitparser.xunit2 import JUnitXml
 
@@ -48,7 +48,7 @@ matches_re_end = re.compile(
 )
 
 
-def _parse_func(file_path: Path):
+def _parse_func(file_path: Path) -> Any:
     from lxml.etree import XMLParser, parse
 
     xml_parser = XMLParser(huge_tree=True)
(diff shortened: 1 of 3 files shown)


KRRT7:
Superseded by #2144 (same commit, rebased onto the linear stack).

PR #2141, opened 2026-05-07
fix: surface subprocess errors when pytest XML is missing

KRRT7:
## Summary

- Surface subprocess stderr/stdout in the error message when pytest exits non-zero and the JUnit XML file is missing
- Previously this was a silent failure with no actionable diagnostics

## Stack

Part of #2132 linear stack.

## Test plan

- [x] Existing test suite passes

### DIFF
--- a/codeflash/languages/python/parse_xml.py
+++ b/codeflash/languages/python/parse_xml.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import os
 import re
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from junitparser.xunit2 import JUnitXml
 
@@ -48,7 +48,7 @@ matches_re_end = re.compile(
 )
 
 
-def _parse_func(file_path: Path):
+def _parse_func(file_path: Path) -> Any:
     from lxml.etree import XMLParser, parse
 
     xml_parser = XMLParser(huge_tree=True)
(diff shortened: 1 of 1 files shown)


codeflash/languages/python/parse_xml.py
```diff
@@ -59,13 +59,22 @@ def parse_python_test_xml(
...
-    run_result: subprocess.CompletedProcess | None = None,
+    run_result: subprocess.CompletedProcess[str] | None = None,
 ) -> TestResults:
     from codeflash.verification.parse_test_output import resolve_test_file_from_class_path
 
     test_results = TestResults()
     if not test_xml_file_path.exists():
-        logger.warning(f"No test results for {test_xml_file_path} found.")
+        if run_result is not None and run_result.returncode != 0:
+            stderr_snippet = (run_result.stderr or "")[:500]
```
aseembits93, line 69: why 500? why not the entire string?

codeflash/languages/python/parse_xml.py
```diff
@@ -234,11 +238,5 @@ def parse_python_test_xml(
...
         )
         if run_result is not None:
-            stdout, stderr = "", ""
-            try:
-                stdout = run_result.stdout.decode()
-                stderr = run_result.stderr.decode()
-            except AttributeError:
-                stdout = run_result.stderr
-            logger.debug(f"Test log - STDOUT : {stdout} \n STDERR : {stderr}")
+            logger.debug(f"Test log - STDOUT : {run_result.stdout} \n STDERR : {run_result.stderr}")
```
aseembits93, line 241: you removed the try except here. are you sure there will be no exceptions?

aseembits93 (approved):

PR #2150, opened 2026-05-09
[FIX] Python stdout comparator when the stdout is empty

mohammedahmed18:


mohammedahmed18:
the e2e-js is fixed inside the mcp PR (https://github.com/codeflash-ai/codeflash/pull/2147)
you can ignore the failing run here

misrasaurabh1 (approved):

PR #2151, opened 2026-05-12
optimize comparator with codeflash-agent autoresearch inspired plugin

mohammedahmed18:
- Chosen benchmark: `.codeflash/benchmarks/test_benchmark_comparator.py::test_benchmark_comparator_rows`
- Correctness oracle: `uv run pytest tests/test_comparator.py -q`
- Baseline metric: `361115 ns`
- Best metric: `111932 ns`
- Overall improvement: `69.00%`
- Total iterations: `2`
- Keeps: `2`
- Discards: `0`
- Best commit: `8548e299`
- Files changed by kept experiments:
  - `codeflash/verification/comparator.py`
- Full comparator benchmark snapshot after the second keep:
  - `rows`: `110976 ns`
  - `nested_dict`: `65314 ns`
  - `identity_types`: `28035 ns`
  - `deep`: `20228 ns`
  - `primitives`: `4701 ns`
- Remaining headroom / blockers:
  - Remaining time is concentrated in the sequence fast-path safety checks inside `comparator()`.
  - More gains are possible, but they likely require narrower special cases or cached shape/type heuristics with higher semantic risk.

### DIFF
--- a/codeflash/verification/comparator.py
+++ b/codeflash/verification/comparator.py
@@ -95,6 +95,8 @@ _IDENTITY_EQ_TYPES: frozenset[type[Any]] = frozenset(
     }
 )
 
+_FAST_SEQUENCE_EQ_TYPES: frozenset[type[Any]] = _IDENTITY_EQ_TYPES | frozenset({float, str})
+
 _EQUALITY_TYPES = (
     int,
     bool,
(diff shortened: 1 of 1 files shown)


mohammedahmed18:
the problem with the plugin now that it's exiting after the first successful optimization, I will update on that issue

PR #2153, opened 2026-05-18
Avoid eager optional comparator imports

aseembits93:
## Summary
- lazily import optional comparator libraries only when comparing objects from those modules
- prevent comparator import from loading pandas/xarray/numexpr during CLI startup
- add a regression test for optional numeric module imports

## Tests
- uv run pytest tests/test_comparator.py -q
- uv run ruff check codeflash/verification/comparator.py --no-fix
- uv run codeflash --help

### DIFF
--- a/codeflash/verification/comparator.py
+++ b/codeflash/verification/comparator.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import _thread
 import array
 import ast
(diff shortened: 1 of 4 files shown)


codeflash/verification/comparator.py
```diff
@@ -117,6 +94,18 @@
 )
 
 
+@cache
+def _optional_module(module_name: str) -> Optional[Any]:
+    if find_spec(module_name) is None:
+        return None
+    return import_module(module_name)
```
KRRT7, line 101: if this is for import time then just remove the entire block and import on demand https://github.com/KRRT7/codeflash/pull/16/changes#diff-a6f2493dbc3f1ec576d03e2d188986f94ca475e3b4cf79ba3dff06309ec01997

PR #2162, opened 2026-05-29
Fix EOFError in non-interactive init flows

mashraf-222:
## What changed
- handle non-interactive `init` confirmation prompts by falling back to safe defaults instead of raising `EOFError`
- skip optional GitHub Actions setup prompts when `init` runs without interactive stdin
- add regression coverage for Python, JavaScript, TypeScript, Java, Go, and the workflow setup path

## Tested
- `uv run pytest -q tests/test_init_yes.py tests/test_init_javascript.py tests/test_init_java_go.py tests/test_github_workflow.py`
- manual plain non-interactive `init` runs with empty stdin in Python, JavaScript, TypeScript, Java, and Go sample repos

### DIFF
--- a/codeflash/cli_cmds/github_workflow.py
+++ b/codeflash/cli_cmds/github_workflow.py
@@ -37,6 +37,7 @@ class DependencyManager(Enum):
 def install_github_actions(override_formatter_check: bool = False, *, skip_confirm: bool = False) -> None:
     try:
         config, _config_file_path = parse_config_file(override_formatter_check=override_formatter_check)
+        interactive_stdin = sys.stdin.isatty()
 
         ph("cli-github-actions-install-started")
         try:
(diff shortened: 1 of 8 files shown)


mashraf-222:
Clarification on the intended behavior here:

- The existing interactive UX is preserved. If `codeflash init` is run from a normal interactive TTY and a valid existing Codeflash config is present, we still show the usual `re-configure?` prompt.
- The change in this PR only affects the broken non-interactive path: if plain `codeflash init` is run without usable stdin, we now fall back to the prompt default instead of surfacing a raw `EOFError` traceback.
- For optional GitHub Actions setup, the non-interactive path skips setup rather than hanging or creating side effects it could not confirm.

So this PR is not removing the existing prompt UX; it only makes the no-stdin path fail safe instead of fail noisy.

mashraf-222:
https://github.com/user-attachments/assets/598f6243-8b56-4184-92ee-5eeef5676b12

